PR #1, opened 2022-11-05
Bootstrap Manila services and provide common operator functions

fmount:
This change adds the basic API definition for manila services, as well as the pkg/ logic to bootstrap the service containers and deploy the related statefulsets.
The templates/ directory is also added and it contains the files needed to prepare the manila environment and push the required config files.
With this we are able to apply a sample CR and deploy the defined pods.


gouthampacha (approved):
Thanks @fmount -- this is a great start. Lets merge as we discussed and continue developing this more

PR #3, opened 2023-02-02
Fix pattern to exclude bin

kajinamit:
Current pattern can match subdirectories.

Signed-off-by: Takashi Kajinami <tkajinam@redhat.com>


fmount (approved):
thanks!

PR #4, opened 2023-02-09
Use time.Duration for timeout

kajinamit:
instead of int. This allows us to use more fine grained timeout value such as 0.1, and encodes the unit in the type itself.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/155
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/166


abays (approved):
/lgtm

Not sure what's going on with the checks though

abays:
@kajinamit Could the linting failures be due to a missing execution of `go mod tidy` before pushing?

kajinamit:
It was my fault. I had to comment out openshift in go.mod to get `go get` happy but I forgot to remove the heading `//` before subimitting PR. I fixed that.

kajinamit:
Hmm. I looked into the lint error, but it seems these are the bugs/problems which are already there in current code it seems. I'll look into these tomorrow when I have time but these should be fixed separately.

abays (approved):
/lgtm

PR #5, opened 2023-02-10
Return error in ConfigMap creation

kajinamit:
The current logic ignores the error but it should be returned so that the error can be detected in the reconciler.


gouthampacha (approved):

gouthampacha:
This looks good, thank you @kajinamit

PR #9, opened 2023-02-14
Fix manila endpoint creation

fmount:
Legacy endpoint shouldn't be used anymore, and we also need to bump the lib-common version to make sure the endpoint are created without appending the port at the end.

Signed-off-by: Francesco Pantano <fpantano@redhat.com>


gouthampacha (approved):
Thanks, LGTM @fmount

PR #12, opened 2023-02-16
Extra volumes support to manila-operator

fmount:
This patch aligns the code with the lib-common ExtraVolumes change, where the ExtraVolumes support is provided for each storage component. Additional extra mounts can be defined in the top level manila CR and passed to the GetVolume functions provided by lib-common. Then, according to the Propagation policy, the extraMount will be (or not be) propagated to the service deployed by manila-operator.

Depends-on: openstack-k8s-operators/lib-common#88


gouthampacha (approved):
Thanks for showing me how this feature works! LGTM @fmount

PR #14, opened 2023-02-20
Switch to time.Duration for timeouts

fmount:
This patch aligns the manila operator to most of the osp-k8s-operators where instead of just using int we switched to time.Duration to express timeouts.


fmount:
Closing because already proposed.

PR #17, opened 2023-02-20
Show overall Manila status in CLI

fmount:
This patch adds the `Status` and `Message` columns so we can print the `status` of the services deployed by manila-operator.


gouthampacha (approved):
Neat, ty @fmount

PR #22, opened 2023-02-22
Fix manifest and keep clusterserviceversion

fmount:
Update the manifest to point to right clusterserviceversion and fix a few references coming from the first import.


config/manifests/bases/manila-operator.clusterserviceversion.yaml
```diff
@@ -25,38 +25,34 @@ spec:
...
     strategy: ""
   installModes:
-  - supported: false
+  - supported: true
     type: OwnNamespace
-  - supported: false
+  - supported: true
     type: SingleNamespace
-  - supported: false
+  - supported: true
```
kajinamit, line 44: I checked the other operators but seems the other operators have MultiNamespace is false (the other options are true)
Should we follow that here ?

kajinamit (changes requested):

config/manifests/bases/manila-operator.clusterserviceversion.yaml
```diff
@@ -25,38 +25,34 @@ spec:
       kind: ManilaScheduler
       name: manilaschedulers.manila.openstack.org
       version: v1beta1
-    - description: ManilaShare is the Schema for the manilashares API
-      displayName: Manila Share
-      kind: ManilaShare
-      name: manilashares.manila.openstack.org
+    - description: ManilaVolume is the Schema for the manilavolumes API
```
kajinamit, line 28: Hmm I think the previous ManilaShare is correct ?

config/manifests/bases/manila-operator.clusterserviceversion.yaml
```diff
@@ -30,33 +30,29 @@ spec:
...
-  - supported: false
+  - supported: true
     type: SingleNamespace
   - supported: false
     type: MultiNamespace
   - supported: true
     type: AllNamespaces
   keywords:
-  - manila
+  - openstack
```
kajinamit, line 49: (this is for follow-up) I understand this is consistent with cinder and glance but the other oeprators are adding different items (for example keystone operator has OpenStack, Identity and Keystone).
We probably have to update cinder/glance/manila operator to follow that standard later.

kajinamit (approved):

config/manifests/bases/manila-operator.clusterserviceversion.yaml
fmount: Sure, let's fix the main issue here and follow up on that minor thing, feel free to propose a patch to align the operators (I'll review!)

PR #23, opened 2023-02-22
Integrate TransportURL when deploying Manila

fmount:
`TransportURL` CR provides a `transport_url` value needed to configure Manila (otherwise manila-share won't start).
`TransportURL` secret is provided by openstack-operator which deploys RabbitMQ and wait for it to be ready: at that point a TransportURL secret is created and can be consumed by Manila.


fmount:
Ok, @gouthampacha I think we should be ready to go here!

controllers/manila_controller.go
```diff
@@ -180,15 +186,63 @@ func (r *ManilaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *ManilaReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// transportURLSecretFn - Watch for changes made to the secret associated with the RabbitMQ
```
gouthampacha, line 189: Thank you for adding this explanation

controllers/manila_controller.go
```diff
@@ -180,15 +186,63 @@ func (r *ManilaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
 		For(&manilav1beta1.Manila{}).
 		Owns(&mariadbv1.MariaDBDatabase{}).
 		Owns(&manilav1beta1.ManilaAPI{}).
 		Owns(&manilav1beta1.ManilaScheduler{}).
 		Owns(&manilav1beta1.ManilaShare{}).
+		Owns(&rabbitmqv1.TransportURL{}).
 		Owns(&batchv1.Job{}).
 		Owns(&corev1.ConfigMap{}).
-		Owns(&corev1.Secret{}).
+		// Watch for TransportURL Secrets which belong to any TransportURLs created by Cinder CRs
```
gouthampacha, line 243: Manila CRs

controllers/manila_controller.go
```diff
@@ -94,6 +98,8 @@ type ManilaReconciler struct {
 // +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;create;update;patch;delete;watch
 // +kubebuilder:rbac:groups=mariadb.openstack.org,resources=mariadbdatabases,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=keystone.openstack.org,resources=keystoneapis,verbs=get;list;watch
+// +kubebuilder:rbac:groups=keystone.openstack.org,resources=keystoneapis,verbs=get;list;watch
```
gouthampacha, line 101: this line is repeated

gouthampacha (approved):
/lgtm 

Thank you for taking me through this code @fmount ; looks great. minor nits inline that we can address in a follow up change. Let's ship it :)

controllers/manila_controller.go
fmount: Done

controllers/manila_controller.go
fmount: good catch thanks!

PR #25, opened 2023-02-23
Run pre-commit with make and add renovate.json

fmount:
As done for other repositories, update pre-commit hook to use make and add renovate.json to handle dependencies.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,19 +1,31 @@
 repos:
-- repo: https://github.com/dnephin/pre-commit-golang
-  rev: v0.5.1
-  hooks:
-    - id: go-fmt
-      exclude: ^vendor
-    - id: go-vet
-    - id: go-mod-tidy
-
-- repo: https://github.com/golangci/golangci-lint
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -1,19 +1,31 @@
...
-    - id: go-mod-tidy
-
-- repo: https://github.com/golangci/golangci-lint
-  rev: v1.50.1
-  hooks:
-    - id: golangci-lint
-
 - repo: local
   hooks:
+    - id: golangci-lint
```
amolkahat, line 4: Need to add tidy and golangci-lint in Makefile[1] 

[1] https://github.com/openstack-k8s-operators/keystone-operator/pull/202

amolkahat (changes requested):

.pre-commit-config.yaml
fmount: Good catch thanks, adding in the next push!

amolkahat (approved):
Thanks @fmount !

PR #27, opened 2023-02-25
Fix rbac, serviceaccount and kubebuilder

fmount:
This patch fixes an issue seen when deploying manila operator using the OLM pattern. kubebuilder is updated to add permissions to the right resources.


fmount:
@gouthampacha this actually fixes the rbac issues in the `manila-operator` bundle.

gouthampacha (approved):

gouthampacha:
Thanks @fmount

PR #33, opened 2023-03-06
Add service db finalizer logic

fmount:
Podified operators which use service dbs and the db.CreateOrPatchDB method , now will get the finalizer set on the db automatically when using the newest lib-common, but they won't have the code remove it on cleanup i.e reconcileDelete, which might make the delete operator action hang. This adds the service Db removal code in the operator delete (reconcileDelete) function to work with the above.

### DIFF
--- a/controllers/manila_controller.go
+++ b/controllers/manila_controller.go
@@ -248,6 +248,18 @@ func (r *ManilaReconciler) SetupWithManager(mgr ctrl.Manager) error {
 func (r *ManilaReconciler) reconcileDelete(ctx context.Context, instance *manilav1beta1.Manila, helper *helper.Helper) (ctrl.Result, error) {
 	r.Log.Info(fmt.Sprintf("Reconciling Service '%s' delete", instance.Name))
 
+	// remove db finalizer first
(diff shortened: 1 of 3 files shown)


fmount:
@gouthampacha merging  this as I need this 1-function change to properly rebase the network-isolation work on top of this.

controllers/manila_controller.go
```diff
@@ -248,6 +248,18 @@ func (r *ManilaReconciler) SetupWithManager(mgr ctrl.Manager) error {
 func (r *ManilaReconciler) reconcileDelete(ctx context.Context, instance *manilav1beta1.Manila, helper *helper.Helper) (ctrl.Result, error) {
 	r.Log.Info(fmt.Sprintf("Reconciling Service '%s' delete", instance.Name))
 
+	// remove db finalizer first
+	db, err := database.GetDatabaseByName(ctx, helper, instance.Name)
```
fmount, line 252: @gouthampacha this represents a sanity check over the db during the delete process and it allows deleting the finalizers set on the resource.

gouthampacha: ack, ty @fmount

PR #34, opened 2023-03-08
Use constant to define restart policy

kajinamit:
... instead hard-coding the string in our own logic.


gouthampacha:
/lgtm 

thanks @kajinamit

gouthampacha (approved):

PR #35, opened 2023-03-08
Add omitempty tag for slices

fmount:
When manila is added to the OpenstackControlPlane, we can observe the job failing with a similar error:

```
The OpenStackControlPlane "openstack" is invalid:

spec.manila.template.manilaScheduler.networkAttachments: Invalid value: "null": spec.manila.template.manilaScheduler.networkAttachments in body must be of type array: "null"
```

This patch adds the "omitempty" tag to the affected arrays specified in the API.

### DIFF
--- a/api/v1beta1/manila_types.go
+++ b/api/v1beta1/manila_types.go
@@ -101,7 +101,7 @@ type ManilaSpec struct {
 
 	// +kubebuilder:validation:Optional
 	// ExtraMounts containing conf files and credentials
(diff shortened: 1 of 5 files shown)


api/v1beta1/manila_types.go
```diff
@@ -101,7 +101,7 @@ type ManilaSpec struct {
```
gouthampacha, line 101: Wouldn't we also want omitempty on other Optional params? (like this one and several above)

gouthampacha (approved):
/lgtm

thanks @fmount

PR #36, opened 2023-03-11
Remove description from Manila CRD

fmount:
This patch rebuilds the manila crd removing the description to avoid a potential issue when the manila bundle is deployed via OLM by the meta operator.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,8 @@ VERIFY_TLS ?= true
 GOWORK ?= off
 export GOWORK := $(GOWORK)
 
+
+GOLINT_TIMEOUT = 100s
 # USE_IMAGE_DIGESTS defines if images are resolved via tags or digests
 # You can enable this value if you would like to use SHA Based Digests
 # To enable set flag to true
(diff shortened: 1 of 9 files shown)


fmount:
@gouthampacha I'm trying to integrate manila-operator in the meta-operator, and I'm currently hitting issues in CI because:

```
time="2023-03-11T07:28:33Z" level=error msg="Bundle files exceeded 1048576 bytes limit" file=/bundle/manifests/nova.openstack.org_nova.yaml
Error: error loading manifests from directory: bundle files exceeded 1048576 bytes limit
Usage:
  opm alpha bundle extract [flags]
```

Basically the total bundle data exceeds the limit (`-l, --datalimit uint  maximum limit in bytes for total bundle data (default 1048576)`.
This might be a bug for the `operator-sdk`, and I see conversations happening upstream, but I'm trying to reduce the amount of data generated by the manila CRD by removing the `description` when `make manifests` is called.
I'm going to merge this change to prove my theory, but we can revert it at any time or just regenerate the `CRD` with the description if needed.

PR #37, opened 2023-03-12
Rename master branch to main

fmount:
We have issues deploying the manila-operator via openstack-operator because the script which is supposed to run the `github action` that pushes the new bundle has "master" instead of "main".
This results in having a "main-latest" [1][2][3] tag instead of just "latest" and the meta-operator is not able to resolve the right manila container! 

[1] https://quay.io/repository/openstack-k8s-operators/manila-operator?tab=tags
[2] https://quay.io/repository/openstack-k8s-operators/manila-operator-index?tab=tags
[3] https://quay.io/repository/openstack-k8s-operators/manila-operator-bundle?tab=tags

### DIFF
--- a/.github/workflows/build-manila-operator.yaml
+++ b/.github/workflows/build-manila-operator.yaml
@@ -35,8 +35,8 @@ jobs:
       id: branch-name
       uses: tj-actions/branch-names@v5
 
-    - name: Set latest tag for non master branch
-      if: "${{ steps.branch-name.outputs.current_branch != 'master' }}"
+    - name: Set latest tag for non main branch
+      if: "${{ steps.branch-name.outputs.current_branch != 'main' }}"
       run: |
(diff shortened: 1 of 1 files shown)


fmount:
@gouthampacha in [1] we point to "latest", but here we have "main-latest" because the script has "master" hardcoded. I'm fixing this in manila-operator operator to keep consistency with the other operators and merging this change as we need this to move forward with the meta-operator integration.

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/213/files#diff-37513468e06c5a876fd5a46e3c193a338a65ab7747195b6d20d575e96ca6e6f0R16

gouthampacha (approved):
/lgtm
/approve

wow, that makes total sense; i saw this tag, but missed this condition in the script! Thanks @fmount

gouthampacha:
https://quay.io/repository/openstack-k8s-operators/manila-operator-bundle?tab=tags&tag=latest

PR #48, opened 2023-04-05
Webhook for default images

abays:
Default container images are now handled via env vars and internal webhooks. This defaulting logic will in turn later be integrated with OpenStack Operator via webhooks there that use the new defaulting mechanism (while at the same time disabling the Manila service operator internal webhooks in that context).


fmount (approved):
Looks good, pretty much same approach we have in Cinder.

fmount:
@abays are you going to follow up here [1] and update the `openstack-operator` when manila is `enabled`?


[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/apis/core/v1beta1/openstackcontrolplane_webhook.go

abays:
> @abays are you going to follow up here [1] and update the `openstack-operator` when manila is `enabled`?
> 
> [1] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/apis/core/v1beta1/openstackcontrolplane_webhook.go

Yes, I will be doing that.  I also have to add logic for all the operators referenced in these files [1], once they themselves have their specific defaulting logic added.  I've been going through that list and adding the webhooks/logic and should hopefully be done soon.

[1] https://github.com/openstack-k8s-operators/openstack-operator/tree/master/pkg/openstack

fmount:
Thanks for the help there, that makes sense!

PR #49, opened 2023-04-05
Enhance liveness probes for manila-operator

silvacarloss:
The manila operator needs to do checks on services status to ensure they are actually up. This change enhances the liveness probes for the manila-share and manila-scheduler services by implementing a HTTP probe using a custom HTTP server. The HTTP server will take advantage of the `manila-manage service list` command, which is an interface manila provides to get the services that are current registered in the database, then allowing the probe to check the services actual state.


gouthampacha (changes requested):
Thanks @silvacarloss. Please see comments inline

templates/manila/bin/healthcheck.py
```diff
@@ -0,0 +1,230 @@
...
+        # Use only some of the fields we know are available in the command 
+        # output. These can change in the future and the change would not
+        # be backwards compatible with this healthcheck, so we keep only 
+        # what is essential.
+        service_headers = ['Binary', 'Host', 'Zone']
+        services_list = services_list.decode('utf-8').split('\n')
+
+        # Alongside the output we want, we have a chance of getting some 
+        # irrelevant things, so we drop that and get the index of where we
+        # should start checking on the output.
```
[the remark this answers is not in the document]
silvacarloss: This is an example of what an output of the command I'm running looks like [1]. The first line is basically an info log
https://paste.opendev.org/show/bYU4HEQ8OYAKhx0nIdLO/
Apparently if I use the --use-journal instead of --nodebug, all info and debug will be redirected, so I made that change and avoided

templates/manila/bin/healthcheck.py
```diff
@@ -0,0 +1,230 @@
...
+
+        # First line is related to the headers. We don't care about the
+        # updated at, so we drop it and sanitize the strings.
+        services_list_headers = services_list[0]
+        columns_to_ignore_starts_at = services_list_headers.index("Updated")
+        services_list_headers = (
+            services_list_headers[0:columns_to_ignore_starts_at])
+        services_list_headers = (
+            [header.lower() for header in services_list_headers])
+        services_list.pop(0)
```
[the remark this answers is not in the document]
silvacarloss: Done - Tried to optimize this, but at the end couldn't get rid of the Updated validation... I think it looks cleaner now

gouthampacha:
/test manila-operator-build-deploy-tempest

think we need to harden the ceph deployment a bit

gouthampacha:
@silvacarloss - can you please rebase this change:

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_manila-operator/49/pull-ci-openstack-k8s-operators-manila-operator-main-manila-operator-build-deploy-tempest/1679261220252684288/artifacts/manila-operator-build-deploy-tempest/openstack-k8s-operators-gather/artifacts/pods/manila-share-share1-0-containers.log

templates/manila/bin/healthcheck.py
```diff
@@ -0,0 +1,227 @@
...
+                 'status': STATUS_ENABLED,
+                 'binary': cls.binary}]
+        else:
+            backend_opts = [
+                cfg.StrOpt('backend_host'),
+                cfg.StrOpt('backend_availability_zone', default=None),
+            ]
+            services_filters = []
+            for backend in CONF.enabled_share_backends:
+                conf = share_conf.Configuration(backend_opts, backend)
```
gouthampacha, line 93: this isn't used anywhere

templates/manila/bin/healthcheck.py
```diff
@@ -0,0 +1,227 @@
...
+        if binary != 'share':
+            services_filters = [
+                {'host': CONF.host,
+                 'state': STATE_UP,
+                 'status': STATUS_ENABLED,
+                 'binary': cls.binary}]
+        else:
+            backend_opts = [
+                cfg.StrOpt('backend_host'),
+                cfg.StrOpt('backend_availability_zone', default=None),
```
gouthampacha, line 89: what are these used for?

templates/manila/bin/healthcheck.py
```diff
@@ -0,0 +1,227 @@
...
+
+        ok = all(service['state'] == STATE_UP for service in our_services)
+        if not ok:
+            raise ServiceNotUp()
+
+    def _parse_service_list_output(self, services_list):
+        # Decode the characters, remove the line splits and the
+        # blank spaces
+        services_list = services_list.decode('utf-8').split('\n')
+        logging.error('DEBUG: services list %s' % services_list)
```
gouthampacha, line 124: don't need this anymore..

templates/manila/bin/healthcheck.py
```diff
@@ -0,0 +1,227 @@
...
+        services_list = [service.split() for service in services_list]
+
+        # First line is related to the headers. We don't care about the
+        # updated at, so we drop it and sanitize the strings.
+        services_list_headers = services_list[0]
+        logging.error(
+            'DEBUG: services list headers %s' % services_list_headers)
```
gouthampacha, line 131: same

gouthampacha (commented):
o/ carlos - just some thoughts inline

templates/manila/bin/healthcheck.py
```diff
@@ -0,0 +1,222 @@
...
+STATE_UP = ":-)"
+STATUS_ENABLED = "enabled"
+
+
+class ServiceNotFoundException(Exception):
+    message = "Service not found."
+
+    def __init__(self, filters={}):
+        if filters:
+            self.message = f'Service not found with filters: {filters}'
```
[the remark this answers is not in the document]
silvacarloss: Done

templates/manila/bin/healthcheck.py
silvacarloss: Yeah, not being used anymore - I remember it was in the past, but I forgot to remove it. Thanks

templates/manila/bin/healthcheck.py
silvacarloss: Done

templates/manila/bin/healthcheck.py
silvacarloss: Done

templates/manila/bin/healthcheck.py
silvacarloss: Done

silvacarloss (commented):
Thanks for the review, please take a look at the most recent changes

silvacarloss:
The kuttl job isn't logging to stdout, and all debug messages are being picked up when we do manila-manage service list. That made me add a workaround to remove the DEBUG info from the manila-manage command output, and I think it could still be relevant, considering that we must treat it in case a customer wants to enable debugging and we don't want the code to break.
The code will look cleaner when the enhancement for manila-manage output format merges upstream.

pkg/manilashare/statefulset.go
```diff
@@ -75,16 +76,17 @@ func StatefulSet(
...
-				"/bin/true",
-			},
+		livenessProbe.HTTPGet = &corev1.HTTPGetAction{
+			Port: intstr.FromInt(8080),
 		}
-		startupProbe.Exec = livenessProbe.Exec
+		startupProbe.HTTPGet = livenessProbe.HTTPGet
+		// Probe doesn't run kolla_set_configs because it uses the 'manila' uid
+		// and gid and doesn't have permissions to make files be owned by root,
+		// so manila.conf is in its original location
```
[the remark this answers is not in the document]
gouthampacha: ^^ this is a minor concern to resolve

templates/manila/bin/healthcheck.py
```diff
@@ -0,0 +1,242 @@
...
+        if not columns_to_ignore_starts_at:
+            raise ServiceNotFoundException(filters={})
+
+        services_list_headers = (
+            services_list_headers[0:columns_to_ignore_starts_at])
+        services_list_headers = (
+            [header.lower() for header in services_list_headers])
+        logging.exception(
+            f'DEBUG: service list headers: {services_list_headers}')
+
```
gouthampacha, line 149: can we drop all these exception logs now? 

iirc we had some trouble getting logs to show up that made us try this route?

gouthampacha (changes requested):
Thanks @silvacarloss ; this looks good.. some minor comments inline.

pkg/manilashare/statefulset.go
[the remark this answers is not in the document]
silvacarloss: Yeah, that's right! Thanks man :)

templates/manila/bin/healthcheck.py
silvacarloss: I added this only to help debugging in the CI, as I was not being able to reproduce this issue in my environment. It can be dropped now :)

templates/manila/bin/healthcheck.py
```diff
@@ -0,0 +1,237 @@
...
+# For share services all enabled backends must be up to return 200, so it is
+# recommended to use a different pod for each backend to avoid one backend
+# affecting others.
+#
+# Requires the name of the service as the first argument (share,
+# scheduler) and optionally a second argument with the location of the
+# configuration directory (defaults to /etc/manila/manila.conf.d)
+
+from http import server
+import logging
```
gouthampacha, line 37: unused import

gouthampacha (approved):
Thanks @silvacarloss 
a minor comment inline; you can drop the unused import in a follow up patch

PR #54, opened 2023-04-21
Disable log file and use stderr for log output

kajinamit:
This change disables usage of log file and makes all log outputs sent to stderr. The file is created inside pod and can fill up container file system.


kajinamit:
Hmm ignore my previous comment. That would be sufficient to disable file logging because we use log_file. Probably we can remove log_dir but the intent is completely different so probably that can be a different PR.

PR #64, opened 2023-05-11
Webhook cleanup

abays:
Cleans up the webhook mechanism based on the latest improvements to the pattern used across our operators.


fmount (approved):
+1 thanks @abays !!

PR #66, opened 2023-05-17
Validate input for Replicas spec

kajinamit:
This adds minimum/maximum value for Replicas spec to avoid accepting a negative number or a too large number. The maximum number follows what is used in keystone-operator.


abays (approved):
/lgtm

PR #67, opened 2023-05-18
Introduce Manila envtest

fmount:
This patch introduces the basic `envTest` structure and defines the common
functions required to simulate the main `Manila` controller reconcile loop.
`lib-common/modules/test` and `lib-common/modules/test-operators` are the
two mod dependencies we rely on to simulate most of the `OpenStack` services
required by Manila.


fmount:
Looks like it works as expected [1]: @gouthampacha @gfidente FYI

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_manila-operator/67/pull-ci-openstack-k8s-operators-manila-operator-main-functional/1659199132843118592/artifacts/test/build-log.txt

api/v1beta1/common_types.go
```diff
@@ -181,3 +182,12 @@ func SetupDefaults() {
...
 }
+
+// GetEnvDefault is a function that gets the value associated with key from environment
+// variables, but use baseDefault as a value in the case of an empty string
+func GetEnvDefault(key string, baseDefault string) string {
+	if value, ok := os.LookupEnv(key); ok {
+		return value
+	}
+	return baseDefault
+}
```
gibizer, line 193: this is already in lib-common https://github.com/openstack-k8s-operators/lib-common/blob/696d321a942e9617c3049d3f1ae0944f20b9ce58/modules/common/util/env.go#LL22C1-L22C1

tests/functional/base_test.go
```diff
@@ -0,0 +1,336 @@
...
+
+func ManilaShareNotExists(name types.NamespacedName) {
+	Consistently(func(g Gomega) {
+		instance := &manilav1.ManilaShare{}
+		err := k8sClient.Get(ctx, name, instance)
+		g.Expect(k8s_errors.IsNotFound(err)).To(BeTrue())
+	}, timeout, interval).Should(Succeed())
+}
+
+func SimulateDeploymentReadyWithPods(name types.NamespacedName, networkIPs map[string][]string) {
```
gibizer, line 243: I think this is a good time to move this to lib-common

tests/functional/base_test.go
```diff
@@ -0,0 +1,336 @@
...
+		ss.Status.Replicas = 1
+		ss.Status.ReadyReplicas = 1
+		g.Expect(k8sClient.Status().Update(ctx, ss)).To(Succeed())
+
+	}, timeout, interval).Should(Succeed())
+
+	logger.Info("Simulated deployment success", "on", name)
+}
+
+func SimulateStatefulSetReplicaReadyWithPods(name types.NamespacedName, networkIPs map[string][]string) {
```
gibizer, line 281: ditto

tests/functional/base_test.go
```diff
@@ -0,0 +1,336 @@
...
+		ss.Status.Replicas = 1
+		ss.Status.ReadyReplicas = 1
+		g.Expect(k8sClient.Status().Update(ctx, ss)).To(Succeed())
+
+	}, timeout, interval).Should(Succeed())
+
+	logger.Info("Simulated statefulset success", "on", name)
+}
+
+func DeleteInstance(instance client.Object) {
```
gibizer, line 319: We already have this in lib-common https://github.com/openstack-k8s-operators/lib-common/blob/696d321a942e9617c3049d3f1ae0944f20b9ce58/modules/test/helpers/instance.go#L25

tests/functional/suite_test.go
```diff
@@ -0,0 +1,228 @@
...
+	Expect(err).NotTo(HaveOccurred())
+	err = networkv1.AddToScheme(scheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	//+kubebuilder:scaffold:scheme
+
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
+	Expect(err).NotTo(HaveOccurred())
+	Expect(k8sClient).NotTo(BeNil())
+	th = NewTestHelper(ctx, k8sClient, timeout, interval, logger)
```
gibizer, line 153: I suggest to pass a specialized logger to the test like 
```
logger.WithName("---Test---")
```
to visually separate the logs coming from the test code from the logs coming from the Reconciler.

gibizer (commented):
This looks like a nice start!

tests/functional/base_test.go
fmount: Agree, and trying to address it here: https://github.com/openstack-k8s-operators/lib-common/pull/274

fmount: I'm going to update this patch as long as the lib-common change lands

tests/functional/base_test.go
fmount: Same of my previous comment.

tests/functional/base_test.go
gibizer: Thank you for pulling this out. I merged the lib-common patch.

tests/functional/base_test.go
```diff
@@ -0,0 +1,239 @@
...
+		"manilaScheduler":  GetDefaultManilaSchedulerSpec(),
+		"manilaShare":      GetDefaultManilaShareSpec(),
+	}
+}
+
+func GetDefaultManilaAPISpec() map[string]interface{} {
+	return map[string]interface{}{
+		"secret":         SecretName,
+		"replicas":       1,
+		"containerImage": manilav1.ManilaAPIContainerImage,
```
gibizer, line 75: If you enable the webhook in suite_test.go then you can drop this and let the webhook default it. Can be done in a follow up.

tests/functional/manila_controller_test.go
```diff
@@ -0,0 +1,387 @@
...
+			Eventually(func() corev1.ConfigMap {
+				return *th.GetConfigMap(manilaTest.ManilaConfigMapScripts)
+			}, timeout, interval).ShouldNot(BeNil())
+		})
+	})
+	When("Manila CR is created without container images defined", func() {
+		BeforeEach(func() {
+			DeferCleanup(th.DeleteInstance, CreateManila(manilaTest.Instance, GetDefaultManilaSpec()))
+		})
+		It("has the expected container image defaults", func() {
```
gibizer, line 203: The current default Specs in base_test contains the container images so this is not really testing any defaulting at the moment. But as I noted there you can enable the webhook in suit_test.go and then drop the images from the specs.

gibizer (approved):
As follow ups:
* I would make this timeout configurable so that we can speed up the test https://github.com/openstack-k8s-operators/manila-operator/blob/main/controllers/manila_controller.go#L364 You can look at nova-operator how we did this there.
* I suggest to switch the test runner from go test to ginkgo. See for example https://github.com
/openstack-k8s-operators/dataplane-operator/pull/204/files This would allow better configuration of test execution
* You can enable webhooks in suite_test.go
* You can add a prow job running the tests based on the job def for nova-operator.

gibizer:
>     * You can add a prow job running the tests based on the job def for nova-operator.
I see this is already there. Nice.

tests/functional/manila_controller_test.go
fmount: Yeah, which is addressed here:  https://github.com/openstack-k8s-operators/manila-operator/pull/77

tests/functional/base_test.go
fmount: Done here: https://github.com/openstack-k8s-operators/manila-operator/pull/77

fmount:
> As follow ups:
> 
>     * I would make this timeout configurable so that we can speed up the test https://github.com/openstack-k8s-operators/manila-operator/blob/main/controllers/manila_controller.go#L364 You can look at nova-operator how we did this there.
> 
>     * I suggest to switch the test runner from go test to ginkgo. See for example https://github.com
>       /[[envtest]Enhance test make target dataplane-operator#204](https://github.com/openstack-k8s-operators/dataplane-operator/pull/204)/files This would allow better configuration of test execution
> 
>     * You can enable webhooks in suite_test.go
> 
>     * You can add a prow job running the tests based on the job def for nova-operator.

I think that the follow up https://github.com/openstack-k8s-operators/manila-operator/pull/77 covers most of your suggestions (Thanks for that).
In particular, in https://github.com/openstack-k8s-operators/manila-operator/pull/77 we:
- switch to the ginkgo cli (re-adapting the `Makefile`)
- Enable webhooks in `suite_test` 
- Fix the `ContainerImage` test

I will follow up on the `timer` being a configurable field, but I'd like to not mix many things in the same change, so I guess I'll take it as yet another action item for later.

PR #78, opened 2023-06-06
Improve NetworkAttachments check

fmount:
Only verify `networkAttachments` if `Replicas` > 0
In addition, `Network` validation is checking the pods if `IPs` are correct
reported for each of the requested networks. For this a label selector is
used to get only the pods for the sub manila component (`api`, `scheduler`
and `share`). Right now there is only the generic `manila app label` on the
deployment resources set. This adds a `component label` on all the `api`,
`scheduler` and `share` deployment resources which allows the check to only
get the wanted resource pods.

### DIFF
--- a/controllers/manila_controller.go
+++ b/controllers/manila_controller.go
@@ -496,14 +496,16 @@ func (r *ManilaReconciler) reconcileNormal(ctx context.Context, instance *manila
 	//
 	// Create ConfigMaps and Secrets required as input for the Service and calculate an overall hash of hashes
 	//
-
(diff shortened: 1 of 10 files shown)


gouthampacha:
/test manila-operator-build-deploy-tempest

the failure here is on teardown of share types (one of the share types failed to get deleted) - seems unrelated

gouthampacha (approved):
/lgtm

tempest must pass since nothing changes for the deployment in the way we're setting this up on the CI

PR #79, opened 2023-06-06
Clean up ManilaShare instances

fmount:
When an entry in the `ManilaShares` spec is deleted, the corresponding `instance` must also be deleted.

**Note** that removing the instance `backend` currently does not clean up the service entry in the `Manila DB`.

That will be handled in a follow-up patch.


fmount:
@gouthampacha @silvacarloss @vhariria FYI this code is tested by kuttl (manila-multibackend) with an additional step where we remove a share before cleanup the top-level CR. By doing this we know this code is executed and produces the expected result.

gouthampacha (approved):
/lgtm 

Thank you; this looks great - the kuttl test is useful too!

PR #85, opened 2023-06-14
Add logging.conf to log to /dev/stdout

fmount:
Due to the recent failures with the new crc/sno versions, Pods are not allowed to log directly in /dev/stdout.
For this reason, like we've done in glance-operator, this patch imports a minimal logging.conf that is referenced in the main config file.

### DIFF
--- a/pkg/manila/dbsync.go
+++ b/pkg/manila/dbsync.go
@@ -79,6 +79,7 @@ func DbSyncJob(instance *manilav1.Manila, labels map[string]string, annotations 
 		UserPasswordSelector: instance.Spec.PasswordSelectors.Service,
 		VolumeMounts:         GetInitVolumeMounts(dbSyncExtraMounts, DbsyncPropagation),
 		Debug:                instance.Spec.Debug.DBInitContainer,
+		LoggingConf:          false,
 	}
 	job.Spec.Template.Spec.InitContainers = InitContainer(initContainerDetails)
 
--- a/pkg/manila/initcontainer.go
+++ b/pkg/manila/initcontainer.go
@@ -16,6 +16,7 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/env"
 
 	corev1 "k8s.io/api/core/v1"
+	"strconv"
 )
 
 // APIDetails information
@@ -31,6 +32,7 @@ type APIDetails struct {
 	VolumeMounts         []corev1.VolumeMount
 	Privileged           bool
 	Debug                bool
+	LoggingConf          bool
 }
 
 const (
@@ -66,6 +68,7 @@ func InitContainer(init APIDetails) []corev1.Container {
 	envVars["DatabaseHost"] = env.SetValue(init.DatabaseHost)
 	envVars["DatabaseUser"] = env.SetValue(init.DatabaseUser)
 	envVars["DatabaseName"] = env.SetValue(init.DatabaseName)
+	envVars["LoggingConf"] = env.SetValue(strconv.FormatBool(init.LoggingConf))
 
 	envs := []corev1.EnvVar{
 		{
--- a/templates/manila/bin/init.sh
+++ b/templates/manila/bin/init.sh
@@ -25,6 +25,7 @@ export DBUSER=${DatabaseUser:-"manila"}
 export DBPASSWORD=${DatabasePassword:?"Please specify a DatabasePassword variable."}
 export PASSWORD=${ManilaPassword:?"Please specify a ManilaPassword variable."}
 export TRANSPORTURL=${TransportURL:-""}
+export LOGGINGCONF=${LoggingConf:-"false"}
 
 export CUSTOMCONF=${CustomConf:-""}
 
@@ -34,6 +35,7 @@ MERGED_DIR=/var/lib/config-data/merged
 SVC_CFG=/etc/manila/manila.conf
 SVC_CFG_MERGED=/var/lib/config-data/merged/manila.conf
 SVC_CFG_MERGED_DIR=${MERGED_DIR}/manila.conf.d
+SVC_CFG_LOGGING=/etc/manila/logging.conf
 
 mkdir -p ${SVC_CFG_MERGED_DIR}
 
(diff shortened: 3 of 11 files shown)


pkg/manila/dbsync.go
```diff
@@ -79,6 +79,7 @@ func DbSyncJob(instance *manilav1.Manila, labels map[string]string, annotations
 		UserPasswordSelector: instance.Spec.PasswordSelectors.Service,
 		VolumeMounts:         GetInitVolumeMounts(dbSyncExtraMounts, DbsyncPropagation),
 		Debug:                instance.Spec.Debug.DBInitContainer,
+		Logging:              false,
```
fmount, line 82: @gouthampacha we don't need logging using the `logging.conf` pattern in `dbsync`

pkg/manila/initcontainer.go
```diff
@@ -66,6 +68,7 @@ func InitContainer(init APIDetails) []corev1.Container {
 	envVars["DatabaseHost"] = env.SetValue(init.DatabaseHost)
 	envVars["DatabaseUser"] = env.SetValue(init.DatabaseUser)
 	envVars["DatabaseName"] = env.SetValue(init.DatabaseName)
+	envVars["Logging"] = env.SetValue(strconv.FormatBool(init.Logging))
```
fmount, line 71: For now, until we remove the `init container` (that will be part of a dedicated patch), the easiest way to define different logging options between `manilaAPI` and `Scheduler` - `Share` is not a conditional in `manila.conf`.
`manila.conf` is inherited by the umbrella `Manila` CR, and it results the very same for all the underlying services.
Instead of making a temporary (complex) change in the Controller(s) logic, it would be a good idea handle this part in the init container with a simple way.
An additional boolean and the related logic in `init.sh` can be used to define which service we want to have the `logging.conf` config option.
By doing this we control where this change will land.

templates/manila/bin/init.sh
```diff
@@ -82,6 +84,14 @@ if [ -f ${CUSTOM_DIR}/custom.conf ]; then
     cp ${CUSTOM_DIR}/custom.conf ${SVC_CFG_MERGED_DIR}/03-service.conf
 fi
 
+if [ "$LOGGING" == "true" ]; then
```
fmount, line 87: @gouthampacha we append that line as part of the `service` config

templates/manila/config/logging.conf
```diff
@@ -0,0 +1,34 @@
+[loggers]
```
fmount, line 1: Imported by glance-operator: @gouthampacha feel free to propose any improvement on this file too!

templates/manila/config/manila-scheduler-config.json
```diff
@@ -6,6 +6,12 @@
       "dest": "/etc/manila/manila.conf.d",
       "owner": "root:manila",
       "perm": "0750"
+    },
+    {
```
fmount, line 10: @gouthampacha we only copy this file on manila-scheduler and manila-share, but not in manila-api that doesn't use this logging mechanism.

pkg/manila/dbsync.go
gouthampacha: Thanks @fmount; I wanted to suggest calling this 

`UsesLoggingConf` or `SupportsLoggingConf`

to avoid confusion.. just a way to clarify that this module doesn't respect logging.conf, but uses logs...

templates/manila/config/logging.conf
gouthampacha: This looks good for now; we'll follow up by lnking debug logging to the log level config on line 25, alongside setting appropriate [library log settings](https://docs.openstack.org/oslo.log/latest/configuration/#DEFAULT.default_log_levels)

templates/manila/config/manila-scheduler-config.json
gouthampacha: ++

gouthampacha (changes requested):

pkg/manila/dbsync.go
fmount: is `LoggingConf` a good solution? It implies a conf file that should exists but don't deny the fact we're logging anyway. If `LoggingConf` doesn't work, I'm ok to change to something like `LoggingConfSupport`, which is basically the same thing.

fmount:
As per my recent experiments, I'm wondering if it makes sense changing this logging model and provide a sidecar container as per the k8s best practices [1].
This change is pretty easy and works very well, however I'd like to put it on hold until we figure out a good logging pattern that can be shared across the board. Looking at [1] seems that sidecars are the recommended way and we can also do log rotation on it without too much effort.
@gouthampacha let's talk more about it, I could probably come up w/ an alternative approach.

[1] https://kubernetes.io/docs/concepts/cluster-administration/logging/

gouthampacha:
> As per my recent experiments, I'm wondering if it makes sense changing this logging model and provide a sidecar container as per the k8s best practices [1]. This change is pretty easy and works very well, however I'd like to put it on hold until we figure out a good logging pattern that can be shared across the board. Looking at [1] seems that sidecars are the recommended way and we can also do log rotation on it without too much effort. @gouthampacha let's talk more about it, I could probably come up w/ an alternative approach.
> 
> [1] https://kubernetes.io/docs/concepts/cluster-administration/logging/

Thanks @fmount - I like this; it's extensible to plug into external logging tools...

fmount:
> > As per my recent experiments, I'm wondering if it makes sense changing this logging model and provide a sidecar container as per the k8s best practices [1]. This change is pretty easy and works very well, however I'd like to put it on hold until we figure out a good logging pattern that can be shared across the board. Looking at [1] seems that sidecars are the recommended way and we can also do log rotation on it without too much effort. @gouthampacha let's talk more about it, I could probably come up w/ an alternative approach.
> > [1] https://kubernetes.io/docs/concepts/cluster-administration/logging/
> 
> Thanks @fmount - I like this; it's extensible to plug into external logging tools...

As per recent conversations, let's move forward with this and we'll move to the side container model as long as we agree on that pattern (e.g. it becomes a requirement) across the board.

gouthampacha (approved):
/lgtm 

Thanks @fmount

PR #86, opened 2023-06-14
Revert "Log to stdout by default"

fmount:
Temporary change until we set proper logging through https://github.com/openstack-k8s-operators/manila-operator/pull/85


fmount:
@gouthampacha merging and the follow up on https://github.com/openstack-k8s-operators/manila-operator/pull/85

PR #88, opened 2023-06-21
Fix manila-api pod name

fmount:
Right now the manila-api pod has no `-api` suffix in its name and it might result confusing for the human operator.
This patch updates the pod name using `instance.Name` where the `-api` suffix is actually present [1].

[1] https://github.com/openstack-k8s-operators/manila-operator/blob/main/controllers/manila_controller.go#L819


gouthampacha:
/lgtm 

Thank you very much @fmount

PR #91, opened 2023-06-23
Stop mouting localtime from host

kajinamit:
Timezone can be configured by TZ environment and this is more flexible. (For example users can configure different timezone for pods and platform.)


gouthampacha:
@kajinamit - sorry for the delay here; needs a rebase now. could you please do that when you get a chance?

PR #93, opened 2023-06-26
Fix inconsistent package name for functional tests

kajinamit:
The package name for functional tests is functional_test instead of functional in most of our operators. Let's use the consistent name.


gouthampacha:
@kajinamit could you please rebase/refresh?

PR #94, opened 2023-06-27
Run make gowork as part of the postUpgrade tasks

fmount:
`precommit-check` seems to fail on multiple renovate related `PRs` due to the fact that `make gowork` is missing. 
By adding this target we can make sure to update both modules in the root of the project and in the `api/`.


abays (approved):
/lgtm

PR #97, opened 2023-07-06
Fix pre-commit in renovate PRs

fmount:
Renovate PRs where a module is updated in both api/ and go.mod present in the root of the projects keep failing because of the missing filter in api/go.{mod,sum} (e.g., see [1])
This patch fixes that behavior and [1] should be unblocked.

[1] https://github.com/openstack-k8s-operators/manila-operator/pull/82


gouthampacha (approved):
/lgtm

PR #99, opened 2023-07-07
Fix Manila deployment with network isolation

fmount:
Somehow the network related resources (NAD) are not registerd in main, hence the deployment will fail if network isolation is enabled.


gouthampacha (approved):
/lgtm

PR #100, opened 2023-07-11
Update test helper references after reorg

gibizer:
Depends-On: openstack-k8s-operators/lib-common#301


gibizer:
The functional failure is reproducible 100% of the time so it is probably related to the bug jump in the lib-common version this PR does.

gibizer:
> The functional failure is reproducible 100% of the time so it is probably related to the big (1,5 months) jump in the lib-common version this PR does.

I see the issue. We need to clean up the initContainers at https://github.com/openstack-k8s-operators/lib-common/blob/54bd95cadcf3772af83e947454f5be8721385fc7/modules/common/test/helpers/deployment.go#L88-L97 similarly how the normal containers are handled there. I will push an enhancement to lib-common

gibizer:
We need to land the PR in lib-common first https://github.com/openstack-k8s-operators/lib-common/pull/302

gibizer:
Should be green now

fmount (approved):

PR #104, opened 2023-07-21
Use central renovate config

fmount:
As done for other operators, this change switches renovate to a central config.


fmount:
@gibizer if you think it's good we can merge w/o waiting for CI as we don't have tests against renovate related changes,

PR #122, opened 2023-08-21
Change kuttl tests to run in arbitrary namespace

vhariria:


vhariria:
recheck

abays:
It looks like something is going wrong in the Ceph kustomize bash script.  Timing out:

```
bash scripts/gen-ceph-kustomize.sh "isready"
~/rh/install_yamls/out/manila-kuttl-tests/ceph/cr ~/rh/install_yamls
Error from server (NotFound): secrets "osp-secret" not found
Waiting the cluster to be up
..............................make: *** [Makefile:1723: ceph] Error 
```

I can reproduce this locally as well.

fmount:
The ceph part is fixed, but I still see issues when manila is applied via kustomize [1]:

```
    logger.go:42: 14:41:57 | deploy/1-deploy-manila | error: no matches for Id Manila.v1beta1.manila.openstack.org/manila.[noNs]; failed to find unique target for patch Manila.v1beta1.manila.openstack.org/manila.[noNs]
    logger.go:42: 14:41:59 | deploy/1-deploy-manila | error: no objects passed to 
```

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_manila-operator/122/pull-ci-openstack-k8s-operators-manila-operator-main-manila-operator-build-deploy-kuttl/1697612815512113152/artifacts/manila-operator-build-deploy-kuttl/openstack-k8s-operators-kuttl/build-log.txt

fmount:
@vhariria ok I think now I managed to get on top of this.
The current patch returns the following error:

```
oc kustomize deployment | oc apply -n $NAMESPACE -f -
error: no matches for Id Manila.v1beta1.manila.openstack.org/manila.[noNs]; failed to find unique target for patch Manila.v1beta1.manila.openstack.org/manila.[noNs]
```

and this seems a side effect of removing the `openstack namespace` that was hardcoded.
I understand we want to deploy on an arbitrary namespace (which is not true given
that we're hardcoding `manila-kuttl-tests` almost everywhere), however I think that
to make things general enough and remove the error above you need to put in both:

- `manila.v1beta.manila.yaml` CR (in config/samples, which is already ok)
- `configure_manila_share_backend.yaml` CR

the same hardcoded namespace (e.g. `namespace: openstack`), and use `kustomize`
to replace this value to the "arbitrary" namespace that is propagated by install_yaml.
If the two files above don't have the same namespace (or it's omitted in one of the CRs),|
the `patchStrategicMerge` will fail with the error above (note that `[noNs]` refers to the
missing namespace that let the patching strategy to fail).
In order to move forward with this patch and prove this theory (that I only tried locally), 
you might want to align the current PR to [1] and see how the CI behaves.

I'll keep an eye on the kuttl execution once there's an update on this patch.

[1] https://github.com/openstack-k8s-operators/manila-operator/compare/main...fmount:manila-operator:kuttl_anyspace

fmount:
@vhariria to unblock a couple of PRs, I'm closing this patch in favor of https://github.com/openstack-k8s-operators/manila-operator/pull/128 that actually fixes the problem. I made a change on top of this one and added you as co-author.

PR #125, opened 2023-08-26
Rework the whole Manila configuration process

fmount:
This patch represents an implementation of the proposal [1] and aligns the `manila-operator` with the work already done in `Cinder` and the other operators.

There are a few relevant changes in the bootstrap process of Manila, in particular:

1. It stops using an `InitContainer` to generate the snippet files that configure each Manila service. The logic that was previously implemented in the `InitContainer` has been moved to the Controllers, where config files are generated and stored in k8s `Secrets`.
2. `InitContainers` are fully removed from the bootstrap process; It uses to copy additional files (httpd and wsgi config in manila-api) to the target directories, and mount the generated config (`0{0,1,2,3}-config.conf` to `/etc/manila/manila.conf.d`, which is used by each service to run;

3. The relevant content, including scripts, previously stored in a `ConfigMap`, are now stored in a corresponding k8s `Secret`, which is mounted to the Service deployment Pods; A total of 4 config snippet files are generated: 
  - 00-config.conf contains global settings that are common to every Manila Pod, including ones that are derived from deployment secrets (e.g. database password, etc.) 
  - 01-config.conf contains the global customServiceConfig settings that apply to every Manila service. 
  - 02-config.conf contains the customServiceConfig settings that are specific to each service. 
  - 03-config.conf contains secrets specified by each service's customServiceConfigSecrets. 

`logging.conf` has been removed as it' s no longer required in the switch to a side container approach for logging purposes, which will be part of a follow up PR.
Finally, functional tests are aligned to the use of k8s `Secrets` instead of the old pattern based on `ConfigMaps`; `kuttl` tests are updated and the `initContainer` has been removed. `DBsync` now mounts only the required files (a minimal `00-config.conf`) and a `db-sync-config.json` containing the command run through `kolla`.

[1] openstack-k8s-operators/docs#31


controllers/manila_controller.go
```diff
@@ -500,12 +499,12 @@ func (r *ManilaReconciler) reconcileNormal(ctx context.Context, instance *manila
 		common.AppSelector: manila.ServiceName,
 	}
 	//
-	// create Configmap required for manila input
+	// create Config required for Nanila input
```
silvacarloss, line 502: Manila

controllers/manilaapi_controller.go
```diff
@@ -223,49 +221,32 @@ func (r *ManilaAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		// CR.Spec.ManagingCrName label matches
+		label := o.GetLabels()
+		if l, ok := label[labels.GetOwnerNameLabelSelector(labels.GetGroupLabel(manila.ServiceName))]; ok {
+			for _, cr := range apis.Items {
+				// return reconcile event for the CR where the owner label AND the parentCinderName matches
+				if l == manila.GetOwningManilaName(&cr) {
+					// return namespace and Name of CR
 					name := client.ObjectKey{
-						Namespace: namespace,
+						Namespace: o.GetNamespace(),
```
silvacarloss, line 233: might be a newbie go question, but wouldn't `Namespace: namespece` be okay too?

controllers/manilascheduler_controller.go
```diff
@@ -558,46 +493,100 @@ func (r *ManilaSchedulerReconciler) reconcileUpgrade(ctx context.Context, instan
 	return ctrl.Result{}, nil
 }
 
-// generateServiceConfigMaps - create custom configmap to hold service-specific config
+// getSecret - get the specified secret, and add its hash to envVars
+func (r *ManilaSchedulerReconciler) getSecret(
```
silvacarloss, line 497: just out of curiosity: is there a way we can reuse this function? Might be another newbie question too :)
Asking because I've realized the only thing changing is *manilav1beta1.ManilaAPI -> *manilav1beta1.ManilaScheduler

silvacarloss (commented):
Looking good @fmount - sorry for the delay on the review.
I have one minor comment and two questions inline, please take a look :)

controllers/manilaapi_controller.go
fmount: This makes sense, yes, I didn't realize while moving / refactoring code that we already assigned `namespace = o.GetNamespace()`: good catch! Changing in my next iteration.

controllers/manilascheduler_controller.go
fmount: so I'm planning to have a follow up on that front, however, I think I have two options there:

1. Use generics and let the function not being tied to any reconciler
2. Create a `BaseReconciler` and let `Manila{API, Scheduler, Share}` do inheritance from it

I see option 2 more realistic and cleaner rather than 1, but I have to experiment with it a bit more, hence targeting for a future improvement when the goal will be doing code cleanup and double check it's stable (mostly Q1 next year).

controllers/manilaapi_controller.go
fmount: Done

controllers/manila_controller.go
fmount: Done

fmount:
> Looking good @fmount - sorry for the delay on the review. I have one minor comment and two questions inline, please take a look :)

Thanks @silvacarloss for the review! Regarding code deduplication, I think we need to follow up on that: this change is big enough and hard to review because of many changes (but all of them are connected together). We'll talk more about it and
we can follow up!

fmount:
The kuttl error here is pretty clear at this point:
- install_yaml switched to the "any-namespace" approach with [1]
- there's a pending patch [2] (which is currently failing) to make the manila operator consistent with the behavior introduced in [1]
- kuttl here is failing because all the resources are created in the `manila-kuttl-test` namespace while the Manila CR is applied against `openstack` namespace. Being `instance.Namespace` == `openstack`, it let the rabbitMQ `transportURL` CR being created in the `openstack` namespace, and `infra` operator isn't able to reconcile it and create the secret, required by Manila to unblock the reconcile loop.


[1]   https://github.com/openstack-k8s-operators/install_yamls/pull/489
[2] https://github.com/openstack-k8s-operators/manila-operator/pull/122

controllers/manilascheduler_controller.go
silvacarloss: Yeah, I think option 2 is a better approach too. Thanks!

silvacarloss (approved):
Looks good, Franceso! Thank you for working on this change

api/v1beta1/common_types.go
```diff
@@ -79,7 +79,7 @@ type ManilaServiceTemplate struct {
 	// +kubebuilder:default="# add your customization here"
 	// CustomServiceConfig - customize the service config using this parameter to change service defaults,
 	// or overwrite rendered information using raw OpenStack config format. The content gets added to
-	// to /etc/<service>/<service>.conf.d directory as custom.conf file.
+	// to /etc/<service>/<service>.conf.d directory a custom config file.
```
gouthampacha, line 82: nit: as a custom config file

gouthampacha (approved):
/lgtm 

thanks @fmount

silvacarloss:
Tardy LGTM. Thanks, Francesco!

PR #127, opened 2023-09-05
Unpin openstack-k8s-operators

kajinamit:
We pinned these to the released versions for the dev preview release but we should resume tracking the latest code.


kajinamit:
Looks like kuttl job is broken and we need to merge https://github.com/openstack-k8s-operators/manila-operator/pull/122 first.

fmount:
> Looks like kuttl job is broken and we need to merge #122 first.

Right, I'll recheck this one as long as the CI is green.

fmount:
@kajinamit you should probably rebase this patch on main to get the kuttl fix

kajinamit:
> @kajinamit you should probably rebase this patch on main to get the kuttl fix

I guess we don't need rebase, but it does not harm, so I rebased this on top on the latest main.

fmount (approved):

PR #130, opened 2023-09-07
Update manilaapi to use service override

stuggi:
Removes creation of routes.Those get done in the openstack-operator. Via service overrides the service can be customized. The service operator adds annotation to the service which needs to be exposed as a route.

Jira: [OSP-26690](https://issues.redhat.com//browse/OSP-26690)

Depends-On: openstack-k8s-operators/lib-common#332


stuggi:
The issue we right now is that the we are building the new operator with the current version of the openstack-operator which has not the functionality to create the route for the new service operator version. With this the public endpoint will be a ClusterIP service and is not reachable by the tempest job which is running outside the OCP env.

I think we have two options:
* update the current samples in the openstack-operator for network-isolation (if that is used by tempest) to also use a LoadBalancer type service for the public endpoint and see if that works for the tempest job
* force merge with failed tempest job result until we merged the openstack-operator PR

stuggi:
> The issue we right now is that the we are building the new operator with the current version of the openstack-operator which has not the functionality to create the route for the new service operator version. With this the public endpoint will be a ClusterIP service and is not reachable by the tempest job which is running outside the OCP env.
> 
> I think we have two options:
> 
> * update the current samples in the openstack-operator for network-isolation (if that is used by tempest) to also use a LoadBalancer type service for the public endpoint and see if that works for the tempest job

This would only work if the host tempest is running is using our DNS server where the service endpoint gets registered

> * force merge with failed tempest job result until we merged the openstack-operator PR

PR #133, opened 2023-09-12
Rename tests directory layout

fmount:
As per the standard golang project layout [1] the directory with the tests should be called `test` and not `tests`, so we rename it here.

We should be consistent in all openstack operators repositories and follow the standard so the documentation has no discrepancies [2].

[1]: https://github.com/golang-standards/project-layout
[2]: https://github.com/openstack-k8s-operators/docs/pull/58/files#r1311828677


fmount:
when we agree to merge this we can follow up on the install_yaml change like we did with both Glance and Cinder [1]

[1] https://github.com/openstack-k8s-operators/install_yamls/pull/506

PR #135, opened 2023-09-15
Modify test folder

vhariria:


fmount:
The good part is that I see now the scale tests running! @gouthampacha I suspect @vhariria found the reason they were not executed!
The bad is that the test is failing [1], looks like we missed something in the `02-assert` but I have to check it better

`Deployment:manila-kuttl-tests/manila-api: .spec.selector.component: key is missing from map` 

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_manila-operator/135/pull-ci-openstack-k8s-operators-manila-operator-main-manila-operator-build-deploy-kuttl/1704586703257210880/artifacts/manila-operator-build-deploy-kuttl/openstack-k8s-operators-kuttl/build-log.txt

vhariria:
@fmount thanks for your feedback. Ack, the failures are resulted from moving the 02-x tests. These tests previously  passed in local environment.  I will update the assert, retest and update the PR as needed.

fmount:
Force merging this patch as it has been already tested locally, CI is green and tempest is expected to be red until the new openstack-operator changes related to the Routes creation land.

PR #139, opened 2023-09-16
Recalculate deployment status

fmount:
If we don't have any `Manila` instance in the deployment, we should mark the deployment condition to False. If `Manila` is enabled in the `OpenStack` control plane, we always need at least 1 replica of API and scheduler, while the `manilaShare` service can be rolled out later in time (according to the backends, hence we should progress even if `ReadyCount` in the `Spec` is zero).
As we've done in `GlanceAPI` through [1], this patch adds the same behavior to the Manila operator.

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/311


abays (approved):
/lgtm

PR #142, opened 2023-09-27
Add manila db purge job

fmount:
Since `Ocata` release [1], `Manila` provides a utility command to purge deleted rows from database [2], and this command can be executed periodically to avoid database tables from getting bigger.
This patch introduces a `Manila` `CronJob` to periodically run `manila-manage db purge`, so that soft-deleted records are removed from `Manila` database. 
The `Cron` scheduling is not exposed through the CR APIs but it's aligned with the default values defined for `TripleO` based deployments.

[1] https://specs.openstack.org/openstack/manila-specs/specs/ocata/add-db-manage-purge.html
[2] https://docs.openstack.org/manila/latest/cli/manila-manage.html


fmount:
we can see here [1] the `manila-cron` job is created. Unfortunately it's still not collected by the `must-gather` tool and I created [2] to make we are able to capture the definition.

```
[pod/manila-operator-controller-manager-6c6d98c949-mmj2t/manager] 2023-09-27T10:34:46Z	INFO	controllers.Manila	CronJob manila-cron - created
```
I tested it locally (by triggering the cronjob manually) and it works as expected.

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_manila-operator/142/pull-ci-openstack-k8s-operators-manila-operator-main-manila-operator-build-deploy-tempest/1706959263483564032/artifacts/manila-operator-build-deploy-tempest/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-070a98749455b097c66a1a25ff827a3c0ee7385f1f0c8bf54f5b15438501f2bd/namespaces/openstack-operators/pods/manila-operator-controller-manager-6c6d98c949-mmj2t/logs/manila-operator-controller-manager-6c6d98c949-mmj2t.log
[2] https://github.com/openstack-k8s-operators/openstack-must-gather/pull/10

fmount:
@gouthampacha @silvacarloss if we manage to get this patch in before I create the `dev-preview2` branch we can get this feature in.

pkg/manila/cronjob.go
```diff
@@ -0,0 +1,124 @@
...
+		{
+			Name: "db-purge-config-data",
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					DefaultMode: &config0644AccessMode,
+					SecretName:  instance.Name + "-config-data",
+					Items: []corev1.KeyToPath{
+						{
+							Key:  DefaultsConfigFileName,
+							Path: DefaultsConfigFileName,
```
silvacarloss, line 62: possibly a dumb question, but I'm assuming this value (DefaultsConfigFileName) is defined elsewhere and we can just consume it, am I correct?

fmount: yeah, it's a `const` defined in the same package and it represents the global config where the service can find the db connection string (which is the only thing we need here).
I only added this config to the cronJob (instead of the whole manila conifig) to limit the mount to what is just required.

silvacarloss: Ah, great! Thanks

vhariria (approved):
Changes to test files lgtm, thanks.

silvacarloss:
tardy lgtm - thanks for the changes, Francesco

PR #147, opened 2023-09-27
Add the new CronJob to the deployed resources that kuttl asserts

fmount:
We are able to run the `DBPurge` `CronJob` using the `--debug` option. 
This PR reworks the way the `DBPurgeCommand` is built, adding the debug option only if the `dbPurge` boolean is set.
In addition, the resulting `CronJob` will be tested at deployment time with an additional kuttl assert.

### DIFF
--- a/test/kuttl/common/assert_sample_deployment.yaml
+++ b/test/kuttl/common/assert_sample_deployment.yaml
@@ -121,6 +121,45 @@ status:
       share1: 1
    transportURLSecret: rabbitmq-transport-url-manila-manila-transport
 ---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: manila-cron
+spec:
+  jobTemplate:
+    metadata:
+      labels:
+        service: manila
+    spec:
+      completions: 1
+      parallelism: 1
+      template:
+        spec:
+          containers:
+          - args:
+            - -c
+            - /usr/bin/manila-manage --config-dir /etc/manila/manila.conf.d
+              db purge 30
+            command:
+            - /bin/bash
+            name: manila-cron
+            volumeMounts:
+            - mountPath: /etc/manila/manila.conf.d
+              name: db-purge-config-data
+              readOnly: true
+          serviceAccount: manila-manila
+          serviceAccountName: manila-manila
+          volumes:
+          - name: db-purge-config-data
+            secret:
+              defaultMode: 420
+              items:
+              - key: 00-config.conf
+                path: 00-config.conf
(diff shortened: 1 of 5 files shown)


test/kuttl/common/assert_sample_deployment.yaml
```diff
@@ -121,6 +121,45 @@ status:
...
+        service: manila
+    spec:
+      completions: 1
+      parallelism: 1
+      template:
+        spec:
+          containers:
+          - args:
+            - -c
+            - /usr/bin/manila-manage --debug --config-dir /etc/manila/manila.conf.d
```
gouthampacha, line 141: good to have this verification ++

gouthampacha: I had this question from the original PR  regarding `--debug`; we're not setting it properly; i.e., if debug for dbSync is True, only "--debug" is set

https://github.com/openstack-k8s-operators/manila-operator/pull/142/files#diff-e37c895a02f32ad01467f37d9743af26a4744a0a77207b25bfdaed6574882778R43-R45

alongside, i wonder if we want to specify a "dbPurge" entry under https://github.com/openstack-k8s-operators/manila-operator/blob/91a921182ce70edf5b5858edf93f0e6a55bc09ed/api/v1beta1/common_types.go#L119-L124

it'd be useful to see the o/p of the pruning from time to time.. 
```
        -            - /usr/bin/manila-manage --debug --config-dir /etc/manila/manila.conf.d
        -              db purge 30
        +            - /usr/bin/manila-manage --config-dir /etc/manila/manila.conf.d db purge
        +              30
```

fmount: Yes, makes sense to decouple the `DBSync` job from the cronjob. I'll do in my next PS (at the beginning I thought that having a single bool to debug all the jobs was fine).
It's better having a separated bool to test/debug the cronjob Pod when it's triggered

fmount:
@vhariria kuttl looks correct: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_manila-operator/147/pull-ci-openstack-k8s-operators-manila-operator-main-manila-operator-build-deploy-kuttl/1707329315756052480/artifacts/manila-operator-build-deploy-kuttl/openstack-k8s-operators-kuttl/build-log.txt

gouthampacha (approved):
Thank you @fmount ; LGTM

PR #152, opened 2023-10-04
Correct order of csv annotation

dprince:
operator-sdk will re-order the csv annotations on local bundle creations This aligns the committed order with what is generated


fmount (approved):
/lgtm

PR #154, opened 2023-10-09
Do not run cronJobs as privileged containers

fmount:
When kolla is not required there's no need to run containers as `privileged` (especially if we run commands that are supposed to interact with DB).
This change removes the root user from the Manila `cronJobs` pod and pass a `SecurityContext` that doesn't violate the `k8s` requirements.


abays (approved):
/lgtm

gouthampacha (approved):
/retest-required

thanks @fmount

PR #155, opened 2023-10-11
Add DHSS related example

fmount:
This patch improves the backends section adding the netapp-dhss one. In addition, it fixes the whole samples by adding the missing bases/openstack sample referenced everywhere.

### DIFF
--- a/config/samples/README.md
+++ b/config/samples/README.md
@@ -16,9 +16,10 @@ credentials in a different yaml file, to having to create secrets.
 
 Currently available samples are:
 
+- native cephfs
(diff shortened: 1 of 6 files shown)


fmount:
We might want to  follow up on the `README` and describe something relevant for each backend. @gouthampacha maybe you can help on that front because we need to specify additional manila commands, but I would do that in a follow up change.

gouthampacha (approved):
great; thanks @fmount .. lgtm. I agree with the readme comment; i'll follow up in a different change

PR #170, opened 2023-10-31
Expose DBPurge Age and Schedule

fmount:
This patch updates the current `manila` `CR` to expose both `dbPurgeAge` and `dbPurgeSchedule` as customizable parameters. They were previously hardcoded as `const` in the code, and it makes sense removing that part as long as the human operator has the ability to customize them.


api/v1beta1/manila_types.go
```diff
@@ -143,6 +147,31 @@ func init() {
...
+// pause in debug mode
+type ManilaDebug struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// dbSync enable debug
+	DBSync bool `json:"dbSync,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// DBPurgeDebug enables debug mode on the DBPurge CronJob
+	DBPurgeDebug bool `json:"debug,omitempty"`
```
fmount, line 172: @ASBishop not sure about your opinion but from a pure CR perspective collecting all the `debug` pieces in a single place helped me to have a better view over the debug area

api/v1beta1/manila_types.go
```diff
@@ -90,6 +90,10 @@ type ManilaSpec struct {
 	// NodeSelector here acts as a default value and can be overridden by service
 	// specific NodeSelector Settings.
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// DBPurge parameters -
+	DBPurgeSpec DBPurge `json:"dbPurge,omitempty"`
```
ASBishop, line 96: Similar to my comment below, I don't think the field's name needs to include the "Spec" suffix, because "Spec" is already part the structure's name. I prefer:

```
	DBPurge DBPurge `json:"dbPurge,omitempty"`
```
The code in cronjob.go would refer to things like "instance.Spec.DBPurge.Age"

api/v1beta1/manila_types.go
```diff
@@ -143,6 +147,31 @@ func init() {
 	SchemeBuilder.Register(&Manila{}, &ManilaList{})
 }
 
+// DBPurge struct is used to model the parameters exposed to the Manila API CronJob
+type DBPurge struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=30
```
ASBishop, line 153: The kuttl failure [1] reveals a problem with initializing the DBPurge structure. 

[1] https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_manila-operator/170/pull-ci-openstack-k8s-operators-manila-operator-main-manila-operator-build-deploy-kuttl/1719630065320333312#1:build-log.txt%3A4201

Even though we want kubebuilder to provide the default values, they aren't set unless the CR explicitly includes dbPurge in the spec:
```
spec:
  dbPurge: {}
```
I did some local tests on the PR code, and changing the kubebuilder validation to Required and/or removing "omitempty" had no effect. I believe this is a side effect of having a defaulting webhook, which makes k8s think the CR has been full initialized, and so it doesn't apply the kubebuilder values.

I don't think it will be sufficient to update *every* manila CR sample to explicitly add the "dbPurge: {}" entry. I think the best way forward is to enhance the webhook code to do something like this:
```
func (spec *ManilaSpec) Default() {
	if (spec.DBPurgeSpec.Age == 0) && (spec.DBPurgeSpec.Schedule == "") {
		spec.DBPurgeSpec.Age = DBPurgeDefaultAge
		spec.DBPurgeSpec.Schedule = DBPurgeDefaultSchedule
	}
```

ASBishop (changes requested):

api/v1beta1/manila_types.go
fmount: ack and thanks for the feedback. 
I see pretty much the same thing looking at the CI execution and I think it's a good practice using webhooks. By doing this we can check (and default) the user input not only when the CR is applied for the first time, but also during the entire lifecycle of the component.
Updating the patch and adding this extra snippet.

api/v1beta1/manila_types.go
fmount: +1 to remove the extra "Spec" suffix

api/v1beta1/common_types.go
```diff
@@ -143,6 +135,8 @@ func SetupDefaults() {
 		APIContainerImageURL:       util.GetEnvVar("RELATED_IMAGE_MANILA_API_IMAGE_URL_DEFAULT", ManilaAPIContainerImage),
 		SchedulerContainerImageURL: util.GetEnvVar("RELATED_IMAGE_MANILA_SCHEDULER_IMAGE_URL_DEFAULT", ManilaSchedulerContainerImage),
 		ShareContainerImageURL:     util.GetEnvVar("RELATED_IMAGE_MANILA_SHARE_IMAGE_URL_DEFAULT", ManilaShareContainerImage),
+		DBPurgeAge:  DBPurgeDefaultAge,
```
fmount, line 138: @ASBishop in the next update I'll change this to get the envVar

fmount:
Let's see the CI in action with this change

fmount:
From the CI execution we can see that the `Manila CR` has the `dbPurge` struct properly initialized [1] and the `cronJob` has been properly created [2].


[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_manila-operator/170/pull-ci-openstack-k8s-operators-manila-operator-main-manila-operator-build-deploy-tempest/1722971224570597376/artifacts/manila-operator-build-deploy-tempest/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-0812f031e363406238d47a3ba3cfb33412e9a2d143d2b4b9365c9796b80bb8aa/namespaces/openstack/crs/manilas.manila.openstack.org/manila.yaml

[2] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_manila-operator/170/pull-ci-openstack-k8s-operators-manila-operator-main-manila-operator-build-deploy-tempest/1722971224570597376/artifacts/manila-operator-build-deploy-tempest/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-0812f031e363406238d47a3ba3cfb33412e9a2d143d2b4b9365c9796b80bb8aa/namespaces/openstack/cronjobs/manila-cron.yaml

ASBishop (changes requested):
In addition to some (very minor) comments noted in the code, I have another suggestion that I'll describe here:

As things currently stand:
- main.go calls manilav1beta1.SetupDefaults()
- But that function is defined in common_types.go, which doesn't feel write to me (it's a function, not a type)
- SetupDefaults (in common_types.go) initializes a local variable with the default values, and then calls SetupManilaDefaults()
- SetupManilaDefaults is implemented in manila_webhook.go, and it essentially saves a copy of the input parameter to the "manilaDefaults" variable defined in the file

This seems overly convoluted. I suggest:
- (Re)move the SetupDefaults() function from common_types.go entirely
- In manila_webhook.go, rename SetupManilaDefaults() to SetupDefaults(), and add the initialization code that was previously done in common_types.go. It would look like this:

```
// SetupDefaults - initialize Manila spec defaults for use with either internal or external webhooks
func SetupDefaults() {
	// Acquire environmental defaults and initialize Manila defaults with them
	manilaDefaults = ManilaDefaults{
		APIContainerImageURL:       util.GetEnvVar("RELATED_IMAGE_MANILA_API_IMAGE_URL_DEFAULT", ManilaAPIContainerImage),
		SchedulerContainerImageURL: util.GetEnvVar("RELATED_IMAGE_MANILA_SCHEDULER_IMAGE_URL_DEFAULT", ManilaSchedulerContainerImage),
		ShareContainerImageURL:     util.GetEnvVar("RELATED_IMAGE_MANILA_SHARE_IMAGE_URL_DEFAULT", ManilaShareContainerImage),
		DBPurgeAge:                 DBPurgeDefaultAge,
		DBPurgeSchedule:            DBPurgeDefaultSchedule,
	}

	manilalog.Info("Manila defaults initialized", "defaults", manilaDefaults)
}
```

ASBishop:
But this is looking good, so I think you can promote if from a mere WIP!

api/v1beta1/common_types.go
```diff
@@ -30,6 +30,11 @@ const (
 	ManilaSchedulerContainerImage = "quay.io/podified-antelope-centos9/openstack-manila-scheduler:current-podified"
 	// ManilaShareContainerImage is the fall-back container image for ManilaShare
 	ManilaShareContainerImage = "quay.io/podified-antelope-centos9/openstack-manila-share:current-podified"
+	//DBPurgeDefaultAge -
```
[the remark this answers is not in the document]
ASBishop: oh, and I think this list of constants might be better located in manila_types.go (they're not common to call controllers, just to top manila controller)

api/v1beta1/manila_types.go
```diff
@@ -143,6 +147,31 @@ func init() {
...
+// pause in debug mode
+type ManilaDebug struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// dbSync enable debug
+	DBSync bool `json:"dbSync,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// DBPurge enables debug mode on the DBPurge CronJob
+	DBPurge bool `json:"debug,omitempty"`
```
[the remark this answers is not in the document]
fmount: +1 it's odd, basically I didn't pay too much attention, I'll fix it in my next PS

api/v1beta1/common_types.go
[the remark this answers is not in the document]
fmount: Yeah, we only have a top-level webhook enforcing the defaults. I agree it makes sense to define them in `manila_types` instead of common, especially because both the `containerImage` and the `dbPurge` fields are top-level fields (and `containerImage` is inherited in the `subCR` controllers).
+1 on this one

fmount:
> In addition to some (very minor) comments noted in the code, I have another suggestion that I'll describe here:
> 
> As things currently stand:
> 
>     * main.go calls manilav1beta1.SetupDefaults()
> 
>     * But that function is defined in common_types.go, which doesn't feel write to me (it's a function, not a type)
> 
>     * SetupDefaults (in common_types.go) initializes a local variable with the default values, and then calls SetupManilaDefaults()
> 
>     * SetupManilaDefaults is implemented in manila_webhook.go, and it essentially saves a copy of the input parameter to the "manilaDefaults" variable defined in the file
> 
> 
> This seems overly convoluted. I suggest:
> 
>     * (Re)move the SetupDefaults() function from common_types.go entirely
> 
>     * In manila_webhook.go, rename SetupManilaDefaults() to SetupDefaults(), and add the initialization code that was previously done in common_types.go. It would look like this:
> 
> 
> ```
> // SetupDefaults - initialize Manila spec defaults for use with either internal or external webhooks
> func SetupDefaults() {
> 	// Acquire environmental defaults and initialize Manila defaults with them
> 	manilaDefaults = ManilaDefaults{
> 		APIContainerImageURL:       util.GetEnvVar("RELATED_IMAGE_MANILA_API_IMAGE_URL_DEFAULT", ManilaAPIContainerImage),
> 		SchedulerContainerImageURL: util.GetEnvVar("RELATED_IMAGE_MANILA_SCHEDULER_IMAGE_URL_DEFAULT", ManilaSchedulerContainerImage),
> 		ShareContainerImageURL:     util.GetEnvVar("RELATED_IMAGE_MANILA_SHARE_IMAGE_URL_DEFAULT", ManilaShareContainerImage),
> 		DBPurgeAge:                 DBPurgeDefaultAge,
> 		DBPurgeSchedule:            DBPurgeDefaultSchedule,
> 	}
> 
> 	manilalog.Info("Manila defaults initialized", "defaults", manilaDefaults)
> }
> ```

This is interesting, thanks for this suggestion that I think is in line w/ the effort around code dedup and help with the whole webhooks related logic.
TBH I didn't pay enough attention at the webhook implementation across the operators (and I tend to disable them in my local testing because they have been used for `containerImages` only), but from a pure coding perspective I see your concern and this seems actually over engineering work ("it wraps the wrapper") that can be simplified.
Let me try what you're suggesting in the next PS and we can do a second pass over the code to see if it's in line with what you have in mind and can be something used consistently across the operators.

fmount:
@viroel I think the zuul kuttl test is still not ready to manage changes in the API, and I'm wondering if we have a way to set it as "non-voting" or override the result.
When it will be stable enough I'll remove the Prow based job.

viroel:
> @viroel I think the zuul kuttl test is still not ready to manage changes in the API, and I'm wondering if we have a way to set it as "non-voting" or override the result. When it will be stable enough I'll remove the Prow based job.

Yes, you just need to set the job to `voting: false`, you can do this at the pipeline [1].
Another issue is that the job is not running due to parent's `files` configuration. You should be able to fix this replacing parent's job with `cifmw-base-multinode-kuttl` instead.

[1] https://github.com/rdo-infra/rdo-jobs/blob/c18955ec6951e50a5cf75b6662c9ab015f42d03c/zuul.d/project-templates.yaml#L206

api/v1beta1/manila_webhook.go
```diff
@@ -82,6 +94,11 @@ func (spec *ManilaSpec) Default() {
 			spec.ManilaShares[key] = manilaShare
 		}
 	}
+
+	if (spec.DBPurge.Age == 0) && (spec.DBPurge.Schedule == "") {
```
ASBishop, line 98: See Gorka's comment for the corresponding code in https://github.com/openstack-k8s-operators/cinder-operator/pull/295

api/v1beta1/manila_types.go
```diff
@@ -148,6 +152,31 @@ func init() {
 	SchemeBuilder.Register(&Manila{}, &ManilaList{})
 }
 
+// DBPurge struct is used to model the parameters exposed to the Manila API CronJob
+type DBPurge struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=30
+	// Age is the DBPurgeAge parameter and indicates the number of days of purging DB records
+	Age int `json:"age"`
```
ASBishop, line 160: Will you add the "+kubebuilder:validation:Minimum=1" in this PR (after CI issues are resolved) or in a follow-up patch?

fmount: I'm going to add this line in a follow up patch because I have that one line change that allows me to focus more on the CI (zuul based) execution rather than the change introduced by the patch.

api/v1beta1/manila_webhook.go
fmount: +1 and thanks for sharing, just commented and updated the patch. I think it's ok to process these two struct fields independently.

ASBishop (approved):
/lgtm

PR #179, opened 2023-11-13
 WIP - Add zuul based kuttl test

fmount:
**Experimenting w/ Zuul**


zuul.d/jobs.yaml
```diff
@@ -0,0 +1,17 @@
+---
+- project:
+    check:
+      jobs:
+        - manila-operator-kuttl
+- job:
+    name: manila-operator-kuttl
+    parent: cifmw-kuttl
```
viroel, line 8: we should go with `cifmw-multinode-kuttl` layout, which is more stable now, the previous one may not be working anymore.

zuul.d/jobs.yaml
```diff
@@ -0,0 +1,17 @@
...
+    attempts: 1
+    vars:
+      cifmw_kuttl_tests_operator_list:
+        - manila
+    required-projects:
```
viroel, line 13: those projects are already defined in parent job, so you may not need to define them here again :)

zuul.d/jobs.yaml
```diff
@@ -0,0 +1,17 @@
+---
+- project:
```
viroel, line 2: create a new "projects.yaml" file with:
```
---
- project:
    name: openstack-k8s-operators/manila-operator
    github-check:
      jobs:
        - manila-operator-kuttl
```

fmount: +1 doing in my next iteration

zuul.d/jobs.yaml
fmount: nice, testing that layout here: https://review.rdoproject.org/r/c/testproject/+/50787
The standalone one works and I'm wondering if it's enough to run kuttl tests, but +1 on moving to that layout as long as it's stable.

zuul.d/jobs.yaml
fmount: +1, removing in my next PS!

fmount (commented):
Thank you @viroel for looking at this patch and suggesting the next steps.
I'm going to update it according to your review and follow up.
FYI testing also via: https://review.rdoproject.org/r/c/testproject/+/50787/

gouthampacha:
/lgtm 

Thanks @fmount; when this merges do we drop the prow kuttl job?

fmount:
I suspect we merged this one too early here as I'm still not sure this job has everything we need to gate patches when we change the API (see https://github.com/openstack-k8s-operators/manila-operator/pull/170)

PR #180, opened 2023-11-14
Remove DefaultConfigOverwrite from CRD

fmount:
As done in Cinder via [1], this patch removes the `DefaultConfigOverwrite` parameter from the `CRD`.
The feature was never fully implemented, and it's superseded by the `ExtraMounts` feature.

[1] https://github.com/openstack-k8s-operators/cinder-operator/pull/297

### DIFF
--- a/api/bases/manila.openstack.org_manilaapis.yaml
+++ b/api/bases/manila.openstack.org_manilaapis.yaml
@@ -60,10 +60,6 @@ spec:
                     default: false
                     type: boolean
                 type: object
-              defaultConfigOverwrite:
-                additionalProperties:
-                  type: string
-                type: object
               extraMounts:
                 items:
                   properties:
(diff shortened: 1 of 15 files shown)


abays (commented):
+1 from me if we've decided to do this across-the-board for all operators

gouthampacha:
> +1 from me if we've decided to do this across-the-board for all operators

+1 from me as well; the discussion elsewhere seemed to indicate a split between wanting this over `extraMounts` in some other operators - because they have a specific use case that we don't have in cinder/manila/glance operators. We support `extraMounts` for a range of purposes already, having this is a small overlap/duplication,  we could document this disparity and merge?

gouthampacha:
/lgtm

This was removed in the cinder-operator via: https://github.com/openstack-k8s-operators/cinder-operator/pull/297

PR #186, opened 2023-11-29
Move -log sidecar container to dumb-init

fmount:
As noticed in other operators, the -log sidecar ignores SIGTERM on delete. This patch moves the manila-operator to the dumb-init usage to make sure the SIGTERM is handled properly by the sidecar.

Fixes: [OSPRH-770](https://issues.redhat.com//browse/OSPRH-770)


gouthampacha (approved):
/lgtm 

thanks @fmount

PR #193, opened 2023-12-13
Move manila-api to StatefulSet

fmount:
As done for many operators, this patch moves `manila-api` to `statefulset` instead of deployment.


ASBishop (approved):
/lgtm

PR #206, opened 2024-01-09
Simplify the way cronjob command base is handled

fmount:
This patch aligns the cronjob code with what it has been done with cinder-operator. In particular the dbPurgeCommand results simplified, with no need of additional string manipulation/conversion.


olliewalsh (approved):
/lgtm

PR #218, opened 2024-01-26
[go.mod]Bump from v0.3.0 to pseudoversion

gibizer:
Our intention is to track service operator and lib-common dependencies
via pseudoversions. However renovate does not automatically bump from a
tagged version (e.g. v0.3.0) to a newer but not tagged pseudoversion. So
this patch does the manual bump. After this renovate will bump the newer
pseduoversions.

Adapted to the new version by:

* adding RBAC rules for MariaDBAccount manipulation
* adding Own(MariaDBAccount) to detect status changes
* adapting test to simulate MariaDBAccount success
* adapting the tests to the interim changes of mariadb TLS


gibizer:
We need a full bump like in nova-operator :/
```
ERROR 1115 (42000) at line 1: Unknown character set: ''
```

gibizer:
> > We need a full bump like in nova-operator :/
> > ```
> > ERROR 1115 (42000) at line 1: Unknown character set: ''
> > ```
> 
> Ah, so I assume you have already seen this error and you're going to follow up with a new PS to make a full bump

Yepp I see this before in nova-operator. We cannot bump to this version of the mariadb-operator in manila either, we need to bump to the latest (including multiple fixes), but that require a list of adaptation in the manila operator as well.  I will try to do the same as I did in https://github.com/openstack-k8s-operators/nova-operator/pull/660

gibizer:
> > > We need a full bump like in nova-operator :/
> > > ```
> > > ERROR 1115 (42000) at line 1: Unknown character set: ''
> > > ```
> > 
> > 
> > Ah, so I assume you have already seen this error and you're going to follow up with a new PS to make a full bump
> 
> Yepp I see this before in nova-operator. We cannot bump to this version of the mariadb-operator in manila either, we need to bump to the latest (including multiple fixes), but that require a list of adaptation in the manila operator as well. I will try to do the same as I did in [openstack-k8s-operators/nova-operator#660](https://github.com/openstack-k8s-operators/nova-operator/pull/660)

pushed the adaptation fixes now. Hopefully CI will agree that this is only what we need to bump to the latest mariadb

fmount (approved):
/lgtm

PR #221, opened 2024-02-02
Do not expose debug section anymore

fmount:
This patch updates the existing API and removes the debug section that was supposed to run the Pod in sleep mode for troubleshooting purposes. This is still possible by using the regular k8s approach, and it's described by [1].

[1] openstack-k8s-operators/docs#90

Implements: [OSPRH-4290](https://issues.redhat.com//browse/OSPRH-4290)


gibizer (approved):
loooks good thanks

PR #224, opened 2024-02-05
Build a FIPS compatible operator image

fmount:
This change is aligned with the work already done for other operators, and the goal is to build a `FIPS` compatible `manila-operator` image. For this reason, the `BASE_IMAGE` has been moved to `ubi-minimal` and the `toolchain` uses `go-toolset`. Both `CGO` and `GO111MODULE` are enabled and passed as extra args to go build.

Implements: [OSPRH-4332](https://issues.redhat.com/browse/OSPRH-4332)

### DIFF
--- a/.prow_ci.env
+++ b/.prow_ci.env
@@ -1,1 +1,2 @@
 export USE_IMAGE_DIGESTS=true
+export FAIL_FIPS_CHECK=true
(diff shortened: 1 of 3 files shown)


abays (approved):
/lgtm

PR #226, opened 2024-02-07
Add the missing fips-compliant annotation to the manifest

fmount:
As per [1] we should have the fips-compliant annotation in the manila-operator clusterserviceversion manifest. This patch adds the missing annotation.

[1] https://docs.openshift.com/container-platform/4.14/operators/operator_sdk/osdk-generating-csvs.html


abays (approved):
/lgtm

PR #229, opened 2024-02-12
Set pod affinity to spread services across OCP workers

ASBishop:
Pod affinity is actually an anti-affinity rule intended to spread pods across all OCP workers. However, we don't want this rule to apply to all manila pods in general. The anti-affinity rule now applies to each specific manila service (api, scheduler, share).


fmount (approved):
/lgtm

PR #242, opened 2024-02-26
[tlse] TLS database connection

stuggi:
The my.cnf file gets added to the secret holding the service configs. The content of my.cnf is centrally managed in the mariadb-operator and retrieved calling db.GetDatabaseClientConfig(tlsCfg)

Depends-On: https://github.com/openstack-k8s-operators/mariadb-operator/pull/190
Depends-On: https://github.com/openstack-k8s-operators/mariadb-operator/pull/191

Jira: [OSPRH-4547](https://issues.redhat.com//browse/OSPRH-4547)

### DIFF
--- a/controllers/manila_controller.go
+++ b/controllers/manila_controller.go
@@ -38,6 +38,7 @@ import (
 	common_rbac "github.com/openstack-k8s-operators/lib-common/modules/common/rbac"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	manilav1beta1 "github.com/openstack-k8s-operators/manila-operator/api/v1beta1"
 	"github.com/openstack-k8s-operators/manila-operator/pkg/manila"
@@ -335,7 +336,7 @@ func (r *ManilaReconciler) reconcileDelete(ctx context.Context, instance *manila
 	r.Log.Info(fmt.Sprintf("Reconciling Service '%s' delete", instance.Name))
 
 	// remove db finalizer first
-	db, err := mariadbv1.GetDatabaseByName(ctx, helper, instance.Name)
+	db, err := mariadbv1.GetDatabaseByName(ctx, helper, manila.DatabaseName)
 	if err != nil && !k8s_errors.IsNotFound(err) {
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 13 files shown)


controllers/manila_controller.go
```diff
@@ -362,65 +363,9 @@ func (r *ManilaReconciler) reconcileInit(
...
-			condition.DBReadyRunningMessage))
-		return ctrlResult, nil
-	}
-	// update Status.DatabaseHostname, used to config the service
-	instance.Status.DatabaseHostname = db.GetDatabaseHostname()
-	instance.Status.Conditions.MarkTrue(condition.DBReadyCondition, condition.DBReadyMessage)
-
+	// TODO: why here?
 	// when job passed, mark NetworkAttachmentsReadyCondition ready
 	instance.Status.Conditions.MarkTrue(condition.NetworkAttachmentsReadyCondition, condition.NetworkAttachmentsReadyMessage)
```
stuggi, line 368: @fmount is this the right place to set the NetworkAttachmentsReadyCondition ?

fmount: @stuggi no! I'm not sure why that line is here. The `NetworkAttachmentsReadyCondition` ready condition should be evaluated in each `sub-level` controller where we call `nad.VerifyNetworkStatusFromAnnotation(` and check the `networkReady` parameter.
Thanks for catching this, I think we can remove this line here.

stuggi: thought so as well, but wanted to double check with you. Removed it

stuggi: @fmount hmm since we initialize the condition and its used to reflect the nad checks at https://github.com/openstack-k8s-operators/manila-operator/blob/main/controllers/manila_controller.go#L642-L669 , I think we could just set it after that block to true in the top level controller.

stuggi: which is basically similar to where it was before as the reconcileInit gets called right after. but its probably cleaner to set it here. let me update the pr

controllers/manila_controller.go
```diff
@@ -590,6 +531,17 @@ func (r *ManilaReconciler) reconcileNormal(ctx context.Context, instance *manila
 	instance.Status.Conditions.MarkTrue(condition.InputReadyCondition, condition.InputReadyMessage)
 	// run check OpenStack secret - end
 
+	//
+	// create service DB instance
+	//
+	db, result, err := r.ensureDB(ctx, helper, instance)
```
fmount, line 537: nit: reuse `ctrlResult` variable, but really a minor thing

fmount (approved):
/lgtm

stuggi:
glance did fail to be up. lets rerun

PR #245, opened 2024-02-28
Add ManilaSpecCore struct

dprince:
This version of the struct (called "core") is meant to be used via the OpenStackControlplane. It is the same as ManilaSpec only it is missing the containerImages from the nested API, Scheduler, and Share specs.

The Default() function for webhooks has been split accordingly.

Jira: [OSPRH-4835](https://issues.redhat.com//browse/OSPRH-4835)


fmount (approved):
/lgtm

fmount:
@dprince  || @abays the change looks good to me, feel free to remove the `hold` label and merge when you're confident w/ that (not sure you're coordinating the change across the board).

PR #257, opened 2024-03-25
Do not add Ceph extraMounts in netapp backend sample

fmount:
ExtraMounts are not required for netapp backend, where a single secret containing the credentials is provided.
It uses the `customServiceConfigSecrets` interface, and it's referenced by name.


silvacarloss (commented):
/lgtm

Thank you for working on this PR.

PR #258, opened 2024-03-26
Fix conditions so they're reset to Unknown at the start of every reconciliation

ASBishop:
This PR follows recent changes to the cinder-operator, which basically initializes conditions to Unknown at the start of the reconcilation loop. As reconciliation progresses, conditions are updated to True or False, and if reconciliation doesn't complete then any remaining conditions are left Unknown.
    
The ReadyCondition is set True when reconciliation successfully completes. For the api, scheduler and share services, this means the number of replicas matches the requested value. Again, this is consistent with the way cinder behaves.
    
For the top manila controller, ReadyCondition is set True when all the other manila controllers are ready.

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -3,7 +3,7 @@ module github.com/openstack-k8s-operators/manila-operator/api
 go 1.20
 
 require (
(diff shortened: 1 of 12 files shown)


fmount (approved):
/lgtm

fmount:
@dprince FYI ^

PR #261, opened 2024-03-29
Mirror ReadyCondition with the actual current condition

fmount:
This patch covers two things:

- `Mirror()` relies on sorting conditions in each condition group by their `LastTransitionTime`, so it swaps the `RestoreLastTransitionTimes()` call with the `Mirror()` call in the defer function

- It setup `ReadyCondition` as `Unknown` instead of `False`, so we can avoid it to be the only one mirrored in
  the instance `Status`: this allows the other conditions to compete for mirroring


abays (approved):
/lgtm

PR #264, opened 2024-04-08
Add ObservedGeneration to the sub Resources

fmount:
This patch does a few things:

- it adds a `Status.OservedGeneration` to the sub custom resources (`API`, `Sched` and `Share`)
- it proposes to bump the `observedGeneration` at the beginning of the reconciliation loop, after we `init` the conditions
- it checks, at the top level, if the `ObservedGeneration` matches with the `metadata.generation` assigned to the `subCR(s)`
- it propagates the full chain to the `StatefulSet`, to make sure we are able to mark `DeploymentReadyCondition` as `True` only when we know the controller is looking at the last `Generation` for the `StatefulSet` as well
-  update the comparison between '.Generation' and '.ObservedGeneration' to match the logic of the [statefulset controller](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/statefulset/stateful_set_utils.go#L574C9-L574C66)


fmount:
For function tests we need https://github.com/openstack-k8s-operators/lib-common/pull/488

controllers/manila_controller.go
```diff
@@ -1126,3 +1136,45 @@ func (r *ManilaReconciler) ensureDB(
 	instance.Status.Conditions.MarkTrue(condition.DBReadyCondition, condition.DBReadyMessage)
 	return db, ctrlResult, nil
 }
+
+// checkManilaSubCRGeneration -
+func (r *ManilaReconciler) checkManilaSubCRGeneration(
```
gibizer, line 1141: This seems like 3 function merged together with the subCR's type erased. I would just have 3 independent functions.

fmount: The goal of a single function is to minimize the duplicate code.
Is your point related to the fact that this function is not clear, or should it be different (like using generics)? 
Or do you see other problems related to the fact I have an `interface{}` in place of the subCR explicit type for each of them?

gibizer: As far as I see the duplicated code it reduces is only
```
listOpts := []client.ListOption{
		client.InNamespace(instance.Namespace),
	}
```
the rest of the current function is different for each type the function handles. 
So I don't think this function is really reduces duplication.

controllers/manila_controller.go
```diff
@@ -676,8 +681,7 @@ func (r *ManilaReconciler) reconcileNormal(ctx context.Context, instance *manila
 		instance.Status.Conditions.Set(c)
 	}
```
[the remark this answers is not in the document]
fmount: Yes, mostly for `Mirror` the condition, `ReadyCount` is assigned and refreshed but I would have an explicit:

```golang
  
if !shareObsGen {
    instance.Status.Conditions.Set(condition.FalseCondition(
	manilav1beta1.ManilaShareReadyCondition,
	condition.ErrorReason,
	condition.SeverityWarning,
	manilav1beta1.ManilaShareReadyErrorMessage,
	err.Error()))
}
```
that also makes easy in the code to see that the `observedGeneration` influences the `DeploymentReady` condition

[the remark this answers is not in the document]
gibizer: That could work. (I would use Unknown instead of False though as it is not a terminal error but something still in progress)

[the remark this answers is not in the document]
fmount: False is in case there's an error while getting the sub resources, but Unknown is ok if `observedGeneration` doesn't match and the result is actually False.

controllers/manila_controller.go
fmount: while technically the single function would work, I think having three separate functions is ok because improves readability and it helps in terms of maintenance, so I switched the implementation to three functions I originally had in the first commit.

controllers/manilashare_controller.go
```diff
@@ -569,16 +570,15 @@ func (r *ManilaShareReconciler) reconcileNormal(ctx context.Context, instance *m
 		return ctrl.Result{}, err
 	}
 
-	if instance.Status.ReadyCount > 0 {
+	if instance.Status.ReadyCount > 0 &&
```
fmount, line 573: @gibizer the `ReadyCount` should be compared w/ the `instance.Spec.Replicas`, but this change will be part of a follow up.

gibizer: OK to handle this as a follow up

controllers/manila_controller.go
```diff
@@ -655,7 +656,24 @@ func (r *ManilaReconciler) reconcileNormal(ctx context.Context, instance *manila
...
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			manilav1beta1.ManilaAPIReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			manilav1beta1.ManilaAPIReadyErrorMessage,
+			err.Error()))
+		return ctrlResult, nil
+	}
+	if !apiObsGen {
```
gibizer, line 669: we need to make the below ready count update and mirroring conditional.

same for the other subCRs below as well

fmount: yes, updating this patch and make sure is aligned with what proposed for designate and octavia

gibizer (approved):
looks good

fmount:
we need https://github.com/openshift/release/pull/51050 to make tempest working again

fmount:
The Prow patch landed, let's retest

PR #265, opened 2024-04-16
Move the sidecar containers to dumb-init

fmount:
As noticed in other operators and for the log sidecar container, `SIGTERM` is ignored on delete. This patch moves the `manila-operator` to entirely use `dumb-init` to make sure the `SIGTERM` is handled properly. In addition, this helps to recreate quickly the containers when tls-e is enabled at `Pod` level.

Jira: [OSPRH-3277](https://issues.redhat.com/browse/OSPRH-3277)


gouthampacha (approved):
/lgtm 

thank you @fmount

PR #269, opened 2024-04-24
Move to golangci-lint-full

fmount:
This patch moves the pre-commit-config to golangci-lint-full and clean up most of the code leftovers.
In particular:
- unused func params are removed
- better use of ctx context.Context
- fix leftovers from previous PRs (e.g., ineff assignments)
- fix revive error for envTests

### DIFF
--- a/controllers/manilascheduler_controller.go
+++ b/controllers/manilascheduler_controller.go
@@ -170,7 +170,7 @@ func (r *ManilaSchedulerReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	// Handle service delete
 	if !instance.DeletionTimestamp.IsZero() {
-		return r.reconcileDelete(ctx, instance, helper)
+		return r.reconcileDelete(instance, helper)
 	}
 
 	// Handle non-deleted clusters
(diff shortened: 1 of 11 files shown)


controllers/manilascheduler_controller.go
```diff
@@ -314,18 +314,6 @@ func (r *ManilaSchedulerReconciler) reconcileDelete(ctx context.Context, instanc
 	return ctrl.Result{}, nil
 }
 
-func (r *ManilaSchedulerReconciler) reconcileInit(
```
fmount, line 317: this function is useless on both scheduler and share, it just logs and return

controllers/manilashare_controller.go
```diff
@@ -314,18 +314,6 @@ func (r *ManilaShareReconciler) reconcileDelete(ctx context.Context, instance *m
 	return ctrl.Result{}, nil
 }
 
-func (r *ManilaShareReconciler) reconcileInit(
```
fmount, line 317: this function is useless and I removed it

gibizer (approved):

PR #274, opened 2024-05-08
Add Scheduler scale up tests

vhariria:


test/kuttl/tests/manila-basic/04-assert.yaml
```diff
@@ -0,0 +1,193 @@
...
+                path: 00-config.conf
+              secretName: manila-config-data
+  schedule: 1 0 * * *
+  suspend: false
+---
+# when using image digests the containerImage URLs are SHA's so we verify them with a script
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  - script: |
```
gibizer, line 166: please add `set -euxo pipefail` to avoid false positives.

vhariria: All set, ty.

PR #285, opened 2024-06-10
Put zuul voting: false

fmount:
`Prow` is still our source of truth until `zuul` is stable enough, hence moving `zuul` to non voting until the switch is completed.


gouthampacha (approved):
/lgtm

Thanks @fmount

PR #288, opened 2024-06-20
Fix api log file increases indefinitely

Akrog:
The manila-api pod uses a file between the manila-api service and a sidecar to pass the manila logs from the service to the stdout.

With the current code it will grow forever, which makes no sense, since the stdout from the pod itself is already being stored, under /var/log/pods and/or in a centralized logging service.

This patch makes the oslo.log service rotate the log file once it reaches 20MB in size.

Other projects like nova and glance set the limit to 50MB and they keep 5 rotated files, but I think that keeping rotated files is useless here and 50MB seems a bit much.

Unfortunately we need to at least keep 1 copy, or oslo.log will not do the right thing if set to 0.

We also need to redirect errors in the sidecar container that tails the log to /dev/null to prevent tail error messages whenever the file gets rotated.

### DIFF
--- a/pkg/manilaapi/statefulset.go
+++ b/pkg/manilaapi/statefulset.go
@@ -128,10 +128,9 @@ func StatefulSet(
 							Args: []string{
 								"--single-child",
 								"--",
-								"/usr/bin/tail",
-								"-n+1",
-								"-F",
-								LogFile,
+								"/bin/sh",
+								"-c",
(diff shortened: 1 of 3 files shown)


fmount:
I think we should update [1] to make kuttl pass.

[1] https://github.com/openstack-k8s-operators/manila-operator/blob/main/test/kuttl/tests/manila-tls/03-assert.yaml#L259

fmount (approved):
/lgtm

PR #292, opened 2024-06-25
Add Required Infrastructure Annotations to CSV

fmount:
Implements: https://issues.redhat.com/browse/OSPRH-7919

### DIFF
--- a/config/manifests/bases/manila-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/manila-operator.clusterserviceversion.yaml
@@ -6,6 +6,11 @@ metadata:
     capabilities: Basic Install
     features.operators.openshift.io/disconnected: "true"
     features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "false"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
     operators.openshift.io/infrastructure-features: '["disconnected"]'
     operators.operatorframework.io/operator-type: non-standalone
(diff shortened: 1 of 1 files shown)


config/manifests/bases/manila-operator.clusterserviceversion.yaml
```diff
@@ -6,6 +6,11 @@ metadata:
     capabilities: Basic Install
     features.operators.openshift.io/disconnected: "true"
     features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "false"
+    features.operators.openshift.io/tls-profiles: "false"
```
fmount, line 10:
```
Whether the operator implements well-known tunables to modify the TLS cipher suite used by the operator and, if applicable, any of the workloads it manages.
```
@gouthampacha wdyt about this?

fmount: Ok looks like the general direction is to set everything to `false` and validate or move to true any relevant parameter later in time.

abays: Yes, that is what I think we're recommending for now

abays (approved):
/lgtm

config/manifests/bases/manila-operator.clusterserviceversion.yaml
stuggi: `operators.openshift.io/infrastructure-features: '["disconnected"]'` bellow is deprecated. we could remove it https://docs.openshift.com/container-platform/4.14/operators/operator_sdk/osdk-generating-csvs.html#osdk-csv-manual-annotations-deprecated_osdk-generating-csvs in a follow up

fmount: ack thank you, and +1 to remove it in a follow up.

PR #316, opened 2024-08-12
Remove errorReason from tls input check functional test

fmount:
We're not really interested in the `ErrorReason` as we depends on what is provided by `lib-common`. However, We still want to match the resulting condition boolean.
This patch removes the `TLSInput` `error message` so we don't fail if something changes in `lib-common`.


abays (approved):
/lgtm

PR #324, opened 2024-09-03
Move to VerifySecret when checking the ctlplane secret

fmount:
Currently we get the provided `osp_secret` and we build the hash of the entire secret.
If an unrelated field is added or changed, this cause a rollout of new manila `Pods`, resulting in an entire service restart in all its components (API, Scheduler and Share services are terminated and recreated).

This change does two main things:
1. replace the `GetSecret` call to `VerifySecret` where possible
2. build a common function that is called by all controllers (top level, `manilaAPI`, `manilaScheduler`, `manilaShare`) to verify the control plane secret exists and has the expected fields

By having a common file for all controllers we are able to also start some code deduplication (that will continue in follow up patches as it's out of scope).

Jira: https://issues.redhat.com/browse/OSPRH-8290

### DIFF
--- a/controllers/manila_common.go
+++ b/controllers/manila_common.go
@@ -38,10 +38,10 @@ type conditionUpdater interface {
 	MarkTrue(t condition.Type, messageFormat string, messageArgs ...interface{})
 }
 
-// getServiceSecret - ensures that the Secret object exists and the expected
+// verifyServiceSecret - ensures that the Secret object exists and the expected
 // fields are in the Secret. It also sets a hash of the values of the expected
 // fields passed as input.
-func getServiceSecret(
+func verifyServiceSecret(
 	ctx context.Context,
 	secretName types.NamespacedName,
 	expectedFields []string,
@@ -73,38 +73,42 @@ func getServiceSecret(
 	return ctrl.Result{}, nil
 }
 
-// getConfigSecret - get the specified secret, and add its hash to envVars
-func getConfigSecret(
+// verifyConfigSecrets - It iterates over the secretNames passed as input and
+// sets the hash of values in the envVars map.
+func verifyConfigSecrets(
 	ctx context.Context,
 	h *helper.Helper,
 	conditionUpdater conditionUpdater,
-	secretName string,
+	secretNames []string,
 	namespace string,
 	envVars *map[string]env.Setter,
 ) (ctrl.Result, error) {
-	secret, hash, err := secret.GetSecret(ctx, h, secretName, namespace)
-	if err != nil {
-		if k8s_errors.IsNotFound(err) {
-			log.FromContext(ctx).Info(fmt.Sprintf("Secret %s not found", secretName))
+	var hash string
+	var err error
+	for _, secretName := range secretNames {
+		_, hash, err = secret.GetSecret(ctx, h, secretName, namespace)
+		if err != nil {
+			if k8s_errors.IsNotFound(err) {
+				log.FromContext(ctx).Info(fmt.Sprintf("Secret %s not found", secretName))
+				conditionUpdater.Set(condition.FalseCondition(
+					condition.InputReadyCondition,
+					condition.RequestedReason,
+					condition.SeverityInfo,
+					condition.InputReadyWaitingMessage))
+				return manila.ResultRequeue, nil
+			}
 			conditionUpdater.Set(condition.FalseCondition(
 				condition.InputReadyCondition,
-				condition.RequestedReason,
-				condition.SeverityInfo,
-				condition.InputReadyWaitingMessage))
-			return manila.ResultRequeue, nil
+				condition.ErrorReason,
+				condition.SeverityWarning,
+				condition.InputReadyErrorMessage,
+				err.Error()))
+			return ctrl.Result{}, err
 		}
-		conditionUpdater.Set(condition.FalseCondition(
-			condition.InputReadyCondition,
-			condition.ErrorReason,
-			condition.SeverityWarning,
-			condition.InputReadyErrorMessage,
-			err.Error()))
-		return ctrl.Result{}, err
+		// Add a prefix to the var name to avoid accidental collision with other non-secret
+		// vars. The secret names themselves will be unique.
+		(*envVars)["secret-"+secretName] = env.SetValue(hash)
 	}
-
-	// Add a prefix to the var name to avoid accidental collision with other non-secret
-	// vars. The secret names themselves will be unique.
-	(*envVars)["secret-"+secret.Name] = env.SetValue(hash)
 
 	return ctrl.Result{}, nil
 }
--- a/controllers/manila_common.go
+++ b/controllers/manila_common.go
@@ -1,0 +1,110 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
+	"k8s.io/apimachinery/pkg/types"
+	"time"
+
+	"github.com/openstack-k8s-operators/lib-common/modules/common/env"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	"github.com/openstack-k8s-operators/manila-operator/pkg/manila"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+type conditionUpdater interface {
+	Set(c *condition.Condition)
+	MarkTrue(t condition.Type, messageFormat string, messageArgs ...interface{})
+}
+
+// getServiceSecret - ensures that the Secret object exists and the expected
+// fields are in the Secret. It also sets a hash of the values of the expected
+// fields passed as input.
+func getServiceSecret(
+	ctx context.Context,
+	secretName types.NamespacedName,
+	expectedFields []string,
+	reader client.Reader,
+	conditionUpdater conditionUpdater,
+	requeueTimeout time.Duration,
+	envVars *map[string]env.Setter,
+) (ctrl.Result, error) {
+
+	hash, res, err := secret.VerifySecret(ctx, secretName, expectedFields, reader, requeueTimeout)
+	if err != nil {
+		conditionUpdater.Set(condition.FalseCondition(
+			condition.InputReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			condition.InputReadyErrorMessage,
+			err.Error()))
+		return res, err
+	} else if (res != ctrl.Result{}) {
+		log.FromContext(ctx).Info(fmt.Sprintf("OpenStack secret %s not found", secretName))
+		conditionUpdater.Set(condition.FalseCondition(
+			condition.InputReadyCondition,
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			condition.InputReadyWaitingMessage))
+		return res, nil
+	}
+	(*envVars)[secretName.Name] = env.SetValue(hash)
+	return ctrl.Result{}, nil
+}
+
+// getConfigSecret - get the specified secret, and add its hash to envVars
+func getConfigSecret(
+	ctx context.Context,
+	h *helper.Helper,
+	conditionUpdater conditionUpdater,
+	secretName string,
+	namespace string,
+	envVars *map[string]env.Setter,
+) (ctrl.Result, error) {
+	secret, hash, err := secret.GetSecret(ctx, h, secretName, namespace)
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
+			log.FromContext(ctx).Info(fmt.Sprintf("Secret %s not found", secretName))
+			conditionUpdater.Set(condition.FalseCondition(
+				condition.InputReadyCondition,
+				condition.RequestedReason,
+				condition.SeverityInfo,
+				condition.InputReadyWaitingMessage))
+			return manila.ResultRequeue, nil
+		}
+		conditionUpdater.Set(condition.FalseCondition(
+			condition.InputReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			condition.InputReadyErrorMessage,
+			err.Error()))
+		return ctrl.Result{}, err
+	}
+
+	// Add a prefix to the var name to avoid accidental collision with other non-secret
+	// vars. The secret names themselves will be unique.
+	(*envVars)["secret-"+secret.Name] = env.SetValue(hash)
+
+	return ctrl.Result{}, nil
+}
--- a/controllers/manila_controller.go
+++ b/controllers/manila_controller.go
@@ -510,7 +510,7 @@ func (r *ManilaReconciler) reconcileNormal(ctx context.Context, instance *manila
 	// check for required OpenStack secret holding passwords for service/admin user and add hash to the vars map
 	//
 
-	result, err := getServiceSecret(
+	result, err := verifyServiceSecret(
 		ctx,
 		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
 		[]string{
--- a/controllers/manila_controller.go
+++ b/controllers/manila_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/go-logr/logr"
 	memcachedv1 "github.com/openstack-k8s-operators/infra-operator/apis/memcached/v1beta1"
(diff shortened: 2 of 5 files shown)


abays:
> @fmount: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> Test name 	Commit 	Details 	Required 	Rerun command
> ci/prow/manila-operator-build-deploy-tempest 	[90945f2](https://github.com/openstack-k8s-operators/manila-operator/commit/90945f2c340497470fbae447f31194e8c9b10ab5) 	[link](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openstack-k8s-operators_manila-operator/324/pull-ci-openstack-k8s-operators-manila-operator-main-manila-operator-build-deploy-tempest/1830914374164811776) 	true 	`/test manila-operator-build-deploy-tempest`
> 
> [Full PR test history](https://prow.ci.openshift.org/pr-history?org=openstack-k8s-operators&repo=manila-operator&pr=324). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:fmount).

```
error: build error: Failed to push image: trying to reuse blob sha256:247c2d03e9487628cb6754ff5385a670df160f7bba36af8fc1f2066e461dc36e at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": net/http: TLS handshake timeout
```

This has been happening in other PRs for other operators as well.  I think something is wrong with Quay.  Retrying just hit the same problem again.  Perhaps we should wait a little while before giving it another shot?

fmount:
> > @fmount: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> > Test name 	Commit 	Details 	Required 	Rerun command
> > ci/prow/manila-operator-build-deploy-tempest 	[90945f2](https://github.com/openstack-k8s-operators/manila-operator/commit/90945f2c340497470fbae447f31194e8c9b10ab5) 	[link](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openstack-k8s-operators_manila-operator/324/pull-ci-openstack-k8s-operators-manila-operator-main-manila-operator-build-deploy-tempest/1830914374164811776) 	true 	`/test manila-operator-build-deploy-tempest`
> > [Full PR test history](https://prow.ci.openshift.org/pr-history?org=openstack-k8s-operators&repo=manila-operator&pr=324). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:fmount).
> 
> ```
> error: build error: Failed to push image: trying to reuse blob sha256:247c2d03e9487628cb6754ff5385a670df160f7bba36af8fc1f2066e461dc36e at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": net/http: TLS handshake timeout
> ```
> 
> This has been happening in other PRs for other operators as well. I think something is wrong with Quay. Retrying just hit the same problem again. Perhaps we should wait a little while before giving it another shot?

I agree, let's wait before rechecking CI

controllers/manilaapi_controller.go
```diff
@@ -638,30 +638,43 @@ func (r *ManilaAPIReconciler) reconcileNormal(ctx context.Context, instance *man
 	//
 	// check for required OpenStack secret holding passwords for service/admin user and add hash to the vars map
 	//
-	ctrlResult, err := r.getSecret(ctx, helper, instance, instance.Spec.TransportURLSecret, &configVars)
+
+	hash, ctrlResult, err := ensureSecret(
```
ASBishop, line 642: Is it necessary to ensureSecret()? I think you should be able to use getConfigSecret().

The reason I say this is the top manila controller doesn't even create the individual deployments until after the TransportURLSecret exists.

controllers/manila_controller.go
```diff
@@ -508,27 +509,21 @@ func (r *ManilaReconciler) reconcileNormal(ctx context.Context, instance *manila
...
-		}
-		instance.Status.Conditions.Set(condition.FalseCondition(
-			condition.InputReadyCondition,
-			condition.ErrorReason,
-			condition.SeverityWarning,
-			condition.InputReadyErrorMessage,
-			err.Error()))
-		return ctrl.Result{}, err
+		return result, err
 	}
```
abays, line 525: Should there be an `else if` here that checks for `result != ctrl.Result{}`, since `ensureSecret` can return a non-empty `ctrl.Result`?

controllers/manilaapi_controller.go
```diff
@@ -638,30 +638,43 @@ func (r *ManilaAPIReconciler) reconcileNormal(ctx context.Context, instance *man
...
+		[]string{
+			"transport_url",
+		},
+		helper.GetClient(),
+		&instance.Status.Conditions,
+		manila.NormalDuration,
+	)
 	if err != nil {
 		return ctrlResult, err
 	}
```
abays, line 654: Same question here about checking for `ctrlResult != ctrl.Result{}`

controllers/manilaapi_controller.go
```diff
@@ -638,30 +638,43 @@ func (r *ManilaAPIReconciler) reconcileNormal(ctx context.Context, instance *man
...
+
 	//
 	// check for required service secrets
 	//
 	for _, secretName := range instance.Spec.CustomServiceConfigSecrets {
-		ctrlResult, err = r.getSecret(ctx, helper, instance, secretName, &configVars)
+		ctrlResult, err = getConfigSecret(ctx, helper, &instance.Status.Conditions, secretName, instance.Namespace, &configVars)
 		if err != nil {
 			return ctrlResult, err
 		}
```
abays, line 667: Same question here about checking for `ctrlResult != ctrl.Result{}`

controllers/manilaapi_controller.go
```diff
@@ -638,30 +638,43 @@ func (r *ManilaAPIReconciler) reconcileNormal(ctx context.Context, instance *man
...
 		fmt.Sprintf("%s-scripts", parentManilaName),     // ScriptsSecret
 		fmt.Sprintf("%s-config-data", parentManilaName), // ConfigSecret
 	}
 
 	for _, parentSecret := range parentSecrets {
-		ctrlResult, err = r.getSecret(ctx, helper, instance, parentSecret, &configVars)
+		ctrlResult, err = getConfigSecret(ctx, helper, &instance.Status.Conditions, parentSecret, instance.Namespace, &configVars)
 		if err != nil {
 			return ctrlResult, err
 		}
```
abays, line 680: Same question here about checking for `ctrlResult != ctrl.Result{}`

controllers/manilaapi_controller.go
fmount: Considering that we don't have to check a particular entry within the `transportURL` secret, I think we can safely use `getConfigSecret()` here. You're right that in theory we're stopped by the top-level controller if transportURL is not ready, and it will create a reconciliation loop.
I'll update this part (same thing applies to scheduler and share) in my next update.

controllers/manila_controller.go
fmount: In theory yes, but in practice `ensureSecret` is a wrapper that calls `VerifySecret` from lib-common, and the extra `else if` that checks `result != ctrl.Result{}` is part of that function (with the `Condition` update as well).
Do you think we should `unwrap` that part and do that here? I thought that it was also a good idea to keep it in the `ensureSecret` function (in manila_common.go) to not duplicate the same code everywhere.

controllers/manilaapi_controller.go
fmount: same answer applies here

controllers/manilaapi_controller.go
fmount: same answer applies here

controllers/manilaapi_controller.go
fmount: same answer applies here

controllers/manilaapi_controller.go
fmount: Done

controllers/manila_controller.go
abays: The problem is that you want to return and stop reconciling here if `result != ctrl.Result{}`.  I agree that the `Condition` is already being properly set, but now we need to kick out of the reconcile loop.  See this [1] for an example of the pattern I'm suggesting.

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/570d08b1749f8139fcf04b7edf2ff905e9fde966/controllers/keystoneapi_controller.go#L718-L734

fmount: Ok, I think we already had this conversation and now I recall we ended up with the check you mentioned in [1] and [2].
The "price" to pay to have that function is this extra `else if` statement to make sure we get out of the loop. 

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/602
[2] https://github.com/openstack-k8s-operators/glance-operator/pull/603

fmount: I think I misread your comment. I was looking at the `ensureSecret` function, that has already the `else if` part, while you were looking at the `getConfigSecret()` function. Now I see your point, updating the patch.

controllers/manilaapi_controller.go
fmount: Done

controllers/manilaapi_controller.go
fmount: Done

controllers/manilaapi_controller.go
fmount: Done

controllers/manila_controller.go
fmount: I think I added the extra check everywhere, including where the `getConfigSecret` function is called.
Let me know if you think I missed something or overlooked any other part.

controllers/manila_common.go
```diff
@@ -0,0 +1,108 @@
...
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			condition.InputReadyWaitingMessage))
+		return "", res, nil
+	}
+	return hash, ctrl.Result{}, nil
+}
+
+// getConfigSecret - get the specified secret, and add its hash to envVars
+func getConfigSecret(
```
ASBishop, line 75: Now that we're using this function for more than just config secrets, perhaps it should be renamed to just getSecret().

controllers/manilashare_controller.go
```diff
@@ -323,26 +323,43 @@ func (r *ManilaShareReconciler) reconcileNormal(ctx context.Context, instance *m
 	//
 	// check for required OpenStack secret holding passwords for service/admin user and add hash to the vars map
 	//
-	ctrlResult, err := r.getSecret(ctx, helper, instance, instance.Spec.Secret, &configVars)
+	hash, ctrlResult, err := ensureSecret(
```
ASBishop, line 326: Another ensureSecret that I don't believe is required.

controllers/manilascheduler_controller.go
```diff
@@ -323,25 +323,44 @@ func (r *ManilaSchedulerReconciler) reconcileNormal(ctx context.Context, instanc
 	//
 	// check for required OpenStack secret holding passwords for service/admin user and add hash to the vars map
 	//
-	ctrlResult, err := r.getSecret(ctx, helper, instance, instance.Spec.Secret, &configVars)
+
+	hash, ctrlResult, err := ensureSecret(
```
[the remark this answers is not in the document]
fmount: I checked the code, and the subCRs do not use `instance.Spec.Secret` other than setting config options referencing the string specified in the CR.
+1 to remove this extra call.

controllers/manilashare_controller.go
fmount: ack and +1 for the same reason of the scheduler code.

controllers/manila_common.go
fmount: Yeah, it was originally called `getSecret()`, I'm ok to keep the original name.

controllers/manilascheduler_controller.go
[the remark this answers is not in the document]
fmount: I think I need to keep this call here and I'll try to explain the reason.
It's true that the top-level controller looks for the existence of `osp-secret` before creating the subCRs, and we're able to block the reconciliation there.
However, when the secret exists and the subCRs (manilaAPI, manilaScheduler, manilaShare) are created, the control moves to the sub-level controller (each resource has a dedicated controller). 
What I do here is ensuring the secret exists, and saving the hash in the status. If the `osp-secret` is updated (because the manila password changes), the goal is to rollout new Pods because the input changed, and ignoring any update to fields not related to `manila`.
Without this check, any subsequent reconciliation loop triggered for the subCR will ignore an update to the secret. For this reason I need to keep this call here.

controllers/manilashare_controller.go
fmount: See the comment I left in the scheduler controller

controllers/manila_common.go
fmount: done

controllers/manilascheduler_controller.go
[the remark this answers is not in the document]
fmount: > I agree we need to keep using ensureSecret(), but I reached the conclusion for different reasons. I'm not sure whether the fact that the services have their own controller matters, but what does matter is they need to leverage ensureSecret()'s ability to return the hash for just the service's own password, and not the entire secret.

Correct, and that hash is what triggers a potential rollout of new Pods is the secret changes. I started from the reconciliation aspect in the previous comment and I was trying to describe the flow, but in the end the hash is what matters for future reconciliation loops.
> 
> With this in mind, since I'm working on the corresponding cinder-operator change, what do you think of renaming it to getServiceSecret() and letting it deal with the expectedFields[] stuff? That might reduce the amount of code that's repeated in the sub-services.

I'm ok to rename `ensureSecret` to `getServiceSecret`, especially if it helps to keep consistency across the storage operators. I'm going to update this change with a couple of more optimizations.
> 
> I expect to have my patch ready tomorrow. It might be worth holding of merging either patch until we iron out all the details, so that we can cross pollinate our ideas.

I'll put this one on /hold, so we can discuss more and converge into a shared idea. Unfortunately I'll be on PTO both Thu and Fri, so we might need to wait until Monday to sync up. Feel free to move forward with your work though, and then we'll discuss more according to your findings.

controllers/manila_common.go
```diff
@@ -0,0 +1,113 @@
...
+			condition.SeverityInfo,
+			condition.InputReadyWaitingMessage))
+		return res, nil
+	}
+	(*envVars)[secretName.Name] = env.SetValue(hash)
+	return ctrl.Result{}, nil
+}
+
+// getConfigSecret - get the specified secret, and add its hash to envVars
+func getConfigSecret(
```
fmount, line 77: @ASBishop I could have unified these two functions. However, I still think that keep them separated (unless you raise concerns) improves readability and we don't have to deal with the difference between `GetSecret` and `VerifySecret` and what they return in terms of error handling. We could still do that in a follow up to focus on the problem that is supposed to solved by this patch.
Let me know your thoughts on this.

controllers/manilaapi_controller.go
```diff
@@ -638,33 +638,45 @@ func (r *ManilaAPIReconciler) reconcileNormal(ctx context.Context, instance *man
...
-		ctrlResult, err = r.getSecret(ctx, helper, instance, secretName, &configVars)
-		if err != nil {
-			return ctrlResult, err
-		}
-	}
 
 	parentManilaName := manila.GetOwningManilaName(instance)
-
-	parentSecrets := []string{
+	secretNames := []string{
```
fmount, line 662: @ASBishop the idea here is to collect the `secretNames` beforehand, including any potential secret referenced by `customServiceConfigSecret`, and make a single call to `getConfigSecret`

controllers/manila_common.go
```diff
@@ -0,0 +1,113 @@
...
+
+type conditionUpdater interface {
+	Set(c *condition.Condition)
+	MarkTrue(t condition.Type, messageFormat string, messageArgs ...interface{})
+}
+
+// getServiceSecret - ensures that the Secret object exists and the expected
+// fields are in the Secret. It also sets a hash of the values of the expected
+// fields passed as input.
+func getServiceSecret(
```
ASBishop, line 44: Per our slack conversation, I recommend naming this function verifyServiceSecret. The reason is is doesn't actually return ("get") the secret. Just like lib-common's VerifySecret() function, it only ensures the secret exists and returns its hash (but not the actual secret, i.e. its value).

controllers/manila_common.go
ASBishop: Yes, I like this approach!

Again, per our slack conversation, I think a better name would be verifyConfigSecrets (plural) for the same reason I mention above.

controllers/manilaapi_controller.go
```diff
@@ -638,33 +638,45 @@ func (r *ManilaAPIReconciler) reconcileNormal(ctx context.Context, instance *man
...
+		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
+		[]string{
+			instance.Spec.PasswordSelectors.Service,
+		},
+		helper.GetClient(),
+		&instance.Status.Conditions,
+		manila.NormalDuration,
+		&configVars,
+	)
+	if (err != nil || ctrlResult != ctrl.Result{}) {
```
ASBishop, line 653: @abays I'm curious if you agree this is equally valid to the "if... else..." logic used elsewhere. I think this compound conditional is more compact, which is a good thing given the fact that this takes place a LOT in our code. I think it would be easier on the eyes if the compound conditional like this could be used.

ASBishop (changes requested):

controllers/manila_common.go
fmount: Makes sense, and done in my last iteration.

controllers/manila_controller.go
```diff
@@ -508,27 +509,23 @@ func (r *ManilaReconciler) reconcileNormal(ctx context.Context, instance *manila
...
-		}
-		instance.Status.Conditions.Set(condition.FalseCondition(
-			condition.InputReadyCondition,
-			condition.ErrorReason,
-			condition.SeverityWarning,
-			condition.InputReadyErrorMessage,
-			err.Error()))
-		return ctrl.Result{}, err
+		return result, err
+	} else if (result != ctrl.Result{}) {
```
ASBishop, line 526: @fmount so I very good with the current patchset , but not sure if you want to use the compound conditional that @abays says is fine. I don't think it's appropriate to mix the two approaches (compound conditional vs if..else), but if people agree the compound variant is cleaner then perhaps a global cleanup would be better in a follow-up patch.

tl;dr I'd be OK merging this as is (once CI completes)

fmount: I agree it requires a cleanup wrt the mixed approach.
I'd be ok to follow up and make more code cleanup and dedup, so if that works and you're happy with the current PS, we can land this patch (if CI allows), and use the new common part as a reference for more code that we can deduplicate.

abays: Is the proposal then to tweak other repos to use the same pattern of a wrapper `verify*Secret()` function around `VerifySecret()` so that in the reconcile loops we can use a compound conditional?  Or are we just talking about making all the conditional in _this_ repo consistent?

ASBishop: I have a couple of orthogonal responses.

The idea of using the compound conditional would apply to any existing code, not just the stuff added for verifying secrets. The idea just happened to come up in this PR.

The storage folks strive to make sure our operators (certainly glance, cinder, manila) are structurally similar. When something affects multiple operators we try to update all of them in as similar manner as possible. I'm working on a PR for the cinder-operator that will be nearly identical to manila's. Perhaps, when we get some time, one of us will submit a PR that replaces multiple if... else... code with the compound conditional, and if that gets accepted then we'll try to repeat it in the other storage operators.

That said, I would not expect EVERY other openstack operator adopt a similar pattern. Consistency across all operators is a great idea, but in practice it's too difficult to get everyone to agree on each pattern. We (storage) try to keep our own houses in order, and it's always nice to see other operators adopt an idea. But that's the extent of our evangelism.

ASBishop: My hot take for cinder is https://github.com/openstack-k8s-operators/cinder-operator/pull/437. It's conceptually similar to this PR, although I organized the commits a little differently.

abays:
I think this looks fine, but I do wonder if we want to homogenize the `if ... else if`s around the `err != nil` and `result != ctrl.Result{}` structures as @ASBishop suggested.  There are a few places [1] [2] [3] where this pattern could be collapsed into the compound approach.

[1] https://github.com/openstack-k8s-operators/manila-operator/blob/9d3f84e95954393ab426e2888c3f0f5d043b2f49/controllers/manila_controller.go#L524-L528
[2] https://github.com/openstack-k8s-operators/manila-operator/blob/9d3f84e95954393ab426e2888c3f0f5d043b2f49/controllers/manilascheduler_controller.go#L338-L342
[3] https://github.com/openstack-k8s-operators/manila-operator/blob/9d3f84e95954393ab426e2888c3f0f5d043b2f49/controllers/manilashare_controller.go#L337-L341

ASBishop:
I think it's OK to merge this as is. There are many more opportunities in other portions of the code to switch to the compound conditional, and that can be handled in a follow up patch. Plus, I know Francesco is OOO for a couple of days.

abays (approved):
/lgtm

PR #336, opened 2024-09-23
Add cronJob envTests

fmount:
This patch improves the existing `envTests` to include `cronJob` verification and update. It shows how an update in the `Manila` top level CR is reflected in an update to the `CronJob` schedule.


abays (approved):
/lgtm

PR #364, opened 2024-11-26
Change test var name to match changes in cifmw

eshulman2:
Changed var name to match the changes in this patch: https://github.com/openstack-k8s-operators/ci-framework/pull/2374/commits


tosky:
Please keep the old variable around and default is value to the value of the new variable. In this way the code won't be broken. Otherwise we would need to merge several changes in a coordinate way with an high risk of breaking something.

PR #371, opened 2024-12-05
Add parameter to customize the number of manilaAPI processes

fmount:


config/crd/bases/manila.openstack.org_manilas.yaml
```diff
@@ -672,6 +672,11 @@ spec:
                       caBundleSecretName:
                         type: string
                     type: object
+                  workers:
+                    default: 3
```
fmount, line 676: @gouthampacha is this a good "default" ? Do we want to stay w/ `4` (the original hardcoded value), or do we want something else?

stuggi:
I am against adding a parameter for something you can use the custom config way to set it. we do not want to introduce tht params

stuggi:
> I am against adding a parameter for something you can use the custom config way to set it. we do not want to introduce tht params

ok, wait this is different. this is for httpd customization

stuggi:
form httpd tuning, check this keystone PR https://github.com/openstack-k8s-operators/keystone-operator/pull/500 . we should do the same here

PR #385, opened 2025-02-03
Consume Topology CR by reference

fmount:
This patch introduces a more granular control over `Pod` placement and `scheduling` through the `Topology` CR integration introduced in [1]. 
In particular it provides:
- a new `API` parameter (`TopologyRef`) defined for each Component that allows to reference an existing `Topology CR` in the same namespace
- the operator logic that retrieves and processes the referenced `Topology CR` through the functions provided by `lib-common` [2]
- an enhanced `StatefulSet` configuration that incorporates the processed `Topology`
- a set of `envTest` to test the lifecycle (add/update/override/remove) of the resulting `StatefulSet` when a `Topology` is referenced

Note that `webhooks` are in place to prevent referencing a `Topology` from a different namespace (which is not a supported scenario).


fmount:
The kuttl failure looks really weird. I think it might be related to some timing issue between multiple set of test execution:

The assert fails with the following:

```
resource Manila:manila-kuttl-tests/manila: .status.conditions.message: value mismatch, expected: Setup complete != actual: DBsync job still running
```
but from the must-gather we can see the DBSync job is over and Pods are up. I'm wondering if we need a bigger timeout to let the assert watch at the last results and move forward.

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_manila-operator/385/pull-ci-openstack-k8s-operators-manila-operator-main-manila-operator-build-deploy-kuttl/1886750025833254912/artifacts/manila-operator-build-deploy-kuttl/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-55eb3b500ace4d56c78d8429339c3bed467f6defa3573ee8de17c6c4a9c1cd7c/namespaces/manila-kuttl-tests/all_resources.log

fmount:
I might need to update this patch on top of https://github.com/openstack-k8s-operators/lib-common/pull/597, /holding it for now.

stuggi:
```
        +  - lastTransitionTime: "2025-02-10T21:45:42Z"
        +    message: DBsync job still running
        +    reason: Requested
        +    severity: Info
        +    status: "False"
```

fmount:
> /retest

@stuggi I analyzed in depth the issue of this kuttl test in [1], and it resulted in [2][3].
The bug in CentOS [3] explains well the whole investigation, and the problem is related to a regression in certificate loading introduced (upstream) in python-requests. The container `current-podified` is affected, while the previous one worked.
Because I'm still not sure how to fix this problem (and btw it happens with all the openstack images, for example I tried neutron), in agreement with the Manila team I might want to override the CI result for now, and keep investigating the issue outside the scope of this patch.
Let me know your thoughts around this, and if we want to approve the patch w/o blacklisting `manila_tls` kuttl.

[1] https://github.com/openstack-k8s-operators/manila-operator/pull/386
[2] https://issues.redhat.com/browse/OSPCIX-673
[3] https://issues.redhat.com/browse/RHEL-78362

stuggi:
> > /retest
> 
> @stuggi I analyzed in depth the issue of this kuttl test in [1], and it resulted in [2][3]. The bug in CentOS [3] explains well the whole investigation, and the problem is related to a regression in certificate loading introduced (upstream) in python-requests. The container `current-podified` is affected, while the previous one worked. Because I'm still not sure how to fix this problem (and btw it happens with all the openstack images, for example I tried neutron), in agreement with the Manila team I might want to override the CI result for now, and keep investigating the issue outside the scope of this patch. Let me know your thoughts around this, and if we want to approve the patch w/o blacklisting `manila_tls` kuttl (that I might do in a different PR until the RHEL issue is fixed and new containers are rebuilt).
> 
> [1] #386 [2] https://issues.redhat.com/browse/OSPCIX-673 [3] https://issues.redhat.com/browse/RHEL-78362

thanks for the links. from my side it sounds good to me to override the check to get it landed if the manila team is also good with it

gouthampacha:
Thanks @fmount; this LGTM.. thanks also for finding and reporting the requests regression. I overrode the job so this unrelated, but important change can merge

PR #390, opened 2025-02-19
Update LastAppliedTopology to TopologyRef interface

fmount:
While `TopologyRef` is able to reference `Name` and `Namespace`, we currently only save `TopologyRef.Name` to `.Status.LastAppliedTopology`.
This patch  aligns both interfaces to use the same type in the same form. In addition, some logic is moved to `infra-operator` because is common to all operators, and this patch also adopts this new form.
It also updates the `SpecCore` `webhook` interface to accept `Namespace` as parameter. It ensures that we properly validate the referenced topology from `openstack-operator`, rejecting requests with a different namespace.

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/1306


config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openstack-k8s-operators/manila-operator
-  newTag: latest
+  newName: quay.io/fpantano/manila-operator
```
fultonj, line 7: we'll want to remove this before merging

fmount: Done

stuggi (approved):
/lgtm

PR #418, opened 2025-05-09
Validate common empty values on CR update

fmount:
We recently hit an issue with the openstack control plane where manila-operator gets stuck waiting for a `transportURL` without progressing or reporting an issue related to an empty `rabbitmqClusterName`.
The same problem applies to all the parameters defined in `ManilaSpecBase(Core)`.
During a CR update, it would be ideal to provide a feedback to the human operator about the not admitted empty parameters, and the goal of this patch is to improve the user experience by introducing base params webhook validations.

Jira: https://issues.redhat.com/browse/OSPRH-16366

### DIFF
--- a/api/v1beta1/manila_webhook.go
+++ b/api/v1beta1/manila_webhook.go
@@ -213,18 +213,25 @@ func (r *Manila) ValidateUpdate(old runtime.Object) (admission.Warnings, error) 
 func (spec *ManilaSpec) ValidateUpdate(old ManilaSpec, basePath *field.Path, namespace string) field.ErrorList {
 	var allErrs field.ErrorList
 
+	// validate the service base parameters
+	allErrs = append(allErrs, spec.ValidateBaseParams(basePath)...)
+
 	// validate the service override key is valid
 	allErrs = append(allErrs, service.ValidateRoutedOverrides(
 		basePath.Child("manilaAPI").Child("override").Child("service"),
 		spec.ManilaAPI.Override.Service)...)
 
 	allErrs = append(allErrs, spec.ValidateManilaTopology(basePath, namespace)...)
+
 	return allErrs
 }
 
 // ValidateUpdate -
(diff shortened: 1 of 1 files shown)


api/v1beta1/manila_webhook.go
```diff
@@ -332,3 +338,45 @@ func (spec *ManilaSpec) ValidateManilaTopology(basePath *field.Path, namespace s
...
+		mcPath := basePath.Child("memcachedInstance")
+		allErrs = append(allErrs,
+			field.Invalid(mcPath, "memcachedInstance", "MemcachedInstance is empty"))
+	}
+	if spec.DatabaseAccount == "" {
+		dbAcc := basePath.Child("databaseAccount")
+		allErrs = append(allErrs,
+			field.Invalid(dbAcc, "databaseAccount", "DatabaseAccount is empty"))
+	}
+	return allErrs
```
gouthampacha, line 381: could refactor this into a method that we can call from both these overloaded forms?

api/v1beta1/manila_webhook.go
```diff
@@ -239,7 +246,6 @@ func (spec *ManilaSpecCore) ValidateUpdate(old ManilaSpecCore, basePath *field.P
 func (r *Manila) ValidateDelete() (admission.Warnings, error) {
 	manilalog.Info("validate delete", "name", r.Name)
```
gouthampacha, line 247: since we have nothing here, i wonder if we need this info log..

gouthampacha (commented):
Hi @fmount, couple of minor questions inline..

api/v1beta1/manila_webhook.go
fmount: I can't atm! Mostly because we need some refactor of the SpecBase to have a single one.
We can plan for this work, but as a dedicated follow up.
(Note that for the same reason I couldn't do it for topologies as well [1])

[1] https://github.com/openstack-k8s-operators/manila-operator/blob/main/api/v1beta1/manila_webhook.go#L274

PR #424, opened 2025-05-26
Do not update status when reconciler exits due to panic

ratailor:
Currently we use deferred call to always update the status field when Reconcile call exits, which seems correct for any other error except Reconciler exits due to panic.

This change adds a call to recover function and try to handle panic and logs error.

Closes: [OSPRH-16933](https://issues.redhat.com//browse/OSPRH-16933)


fmount (approved):
/lgtm

PR #428, opened 2025-06-06
Enable MTLS memcached auth

lmiccini:
This commit allows operators to use mtls as an authentication method against Memcached.
Manila controllers will detect the presence of a purposely-created mtls certificate (authCertSecret) and use this to configure the [keystone_authtoken] section accordingly.
Additional volumes/volumemounts will be appended to each pod.

Note that this commit switches from MemcachedServersWithInet to MemcachedServers since keystone-middleware now uses pymemcache when tls=true and there is no need to use "[]" to enclose the list of memcached servers even for ipv6.


stuggi (approved):
/lgtm

lmiccini:
time="2025-07-10T10:29:04Z" level=fatal msg="Error generating bundle manifests: error resolving image: GET https://quay.io/v2/auth?scope=repository%3Apodified-antelope-centos9%2Fopenstack-designate-backend-bind9%3Apull&service=quay.io: unexpected status code 504 Gateway Timeout:

PR #433, opened 2025-06-23
Enable notifications for Manila

fmount:
This patch introduces a new `notificationInstanceBus` parameter in the `Manila` CR. This parameter is used to create a `TransportURL` that will be used for notification purposes. When a `notificationInstanceBus` parameter is set, manila controller:

1. init the `Notification` condition provided by `lib-common`
2. in case of a dedicated `rabbitmq` instance, requests a new `TransportURL` following the known pattern
3. gets from the resulting k8s `Secret` the info required to customize `00-config.conf`, where `oslo_messaging_notifications` is present 

Jira: https://issues.redhat.com/browse/OSPRH-17699


api/v1beta1/manila_types.go
```diff
@@ -30,6 +30,14 @@ const (
...
 	// DeploymentHash hash used to detect changes
 	DeploymentHash = "deployment"
+	// MessagingBus - TransportURLSecret messagingBus
+	MessagingBus = "messagingBus"
+	// NotificationBus - TransportURLSecret notifications
+	NotificationBus = "notificationBus"
+	// NotificationDriverNoop -
+	NotificationDriverNoop = "noop"
+	// NotificationDriverMessagingv2 -
+	NotificationDriverMessagingv2 = "messagingv2"
```
stuggi, line 40: its not super critical, but do we need those 2 driver const? could just handle all in the template

```
[oslo_messaging_notifications]
{{ if (index . "NotificationsURL") -}}
transport_url = {{ .NotificationsURL }}
driver = messagingv2
{{ else }}
driver = noop
{{ end -}}
```

stuggi: yes, that was basically my idea. seems thats how it was also done in the nova-op to have the condition in the template https://github.com/openstack-k8s-operators/nova-operator/blob/main/templates/nova.conf#L139-L145

fmount: ok I will switch to that approach in my next PS

controllers/manila_controller.go
```diff
@@ -197,6 +197,14 @@ func (r *ManilaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 	instance.Status.Conditions.Init(&cl)
 	instance.Status.ObservedGeneration = instance.Generation
 
+	if instance.Spec.NotificationBusInstance != nil {
```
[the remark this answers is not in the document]
fmount: done

api/v1beta1/manila_types.go
```diff
@@ -142,7 +151,7 @@ type ManilaStatus struct {
 	DatabaseHostname string `json:"databaseHostname,omitempty"`
 
 	// TransportURLSecret - Secret containing RabbitMQ transportURL
-	TransportURLSecret string `json:"transportURLSecret,omitempty"`
+	TransportURLSecret map[string]string `json:"transportURLSecret,omitempty"`
```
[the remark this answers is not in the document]
fmount: Let me locally test a minor update. If I see an issue in the transition between the old and the new operator I will change this part to have a dedicated `NotificationURLSecret` in the `.Status`.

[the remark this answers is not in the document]
stuggi: as said, I think it can work with how we apply the CRDs today. if we'd still use OLM, I think it would block installing the new CRD on a non greenfield deployment. But yes, good to give it a try.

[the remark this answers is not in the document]
fmount: ok, I tested the minor update flow as described in my previous comment with the last change and it looks good. I was able to see that everything looks fine and I can successfully manage the lifecycle of this variable.
@stuggi if you can take a look at the remaining open points we can probably move forward with this! Thank you for the good pointers so far.

controllers/manila_controller.go
```diff
@@ -957,6 +1016,21 @@ func (r *ManilaReconciler) generateServiceConfig(
 	}
 	templateParameters["VHosts"] = httpdVhostConfig
 
+	var notificationInstanceURLSecret *corev1.Secret
+	if instance.Status.NotificationURLSecret != nil {
+		// Get a notificationInstanceURLSecret only if rabbitMQ referenced in
+		// the spec is different, otherwise inherits the existing transport_url
+		if instance.Spec.RabbitMqClusterName != *instance.Spec.NotificationBusInstance {
```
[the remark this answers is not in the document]
stuggi: can `*instance.Spec.NotificationBusInstance` be nil here?

controllers/manila_controller.go
```diff
@@ -521,6 +530,56 @@ func (r *ManilaReconciler) reconcileNormal(ctx context.Context, instance *manila
...
+	//
+
+	// Request TransportURL when the parameter is provided in the CR
+	// and it does not match with the existing RabbitMqClusterName
+	if instance.Spec.NotificationBusInstance != nil {
+
+		// init .Status.NotificationURLSecret
+		if instance.Status.NotificationURLSecret == nil {
+			instance.Status.NotificationURLSecret = ptr.To("")
+		}
```
stuggi, line 545: I guess we could remove this init if we just use this bellow? `		instance.Status.NotificationURLSecret = ptr.To(notificationBusInstanceURL.Status.SecretName)`

controllers/manila_controller.go
```diff
@@ -1142,16 +1224,20 @@ func (r *ManilaReconciler) shareDeploymentCreateOrUpdate(
...
-func (r *ManilaReconciler) transportURLCreateOrUpdate(ctx context.Context, instance *manilav1beta1.Manila) (*rabbitmqv1.TransportURL, controllerutil.OperationResult, error) {
+func (r *ManilaReconciler) transportURLCreateOrUpdate(
+	ctx context.Context,
+	instance *manilav1beta1.Manila,
+	rabbitMqClusterName string,
+) (*rabbitmqv1.TransportURL, controllerutil.OperationResult, error) {
 	transportURL := &rabbitmqv1.TransportURL{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-manila-transport", instance.Name),
+			Name:      fmt.Sprintf("%s-manila-transport-%s", instance.Name, rabbitMqClusterName),
```
stuggi, line 1234: For an env update, this will result in having a new transporturl, but still the old with without the suffix. we probably have to either cleanup the old one, or pass an empty string for the RabbitMqClusterName so we only use a suffix for a dedicated NotificationBusInstance.

controllers/manila_controller.go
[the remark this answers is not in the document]
fmount: no because when the trasportURL is created, `.Status.NotifcationURLSecret` is set only when `instance.Spec.NotificationBusInstance` is not `nil`, so that early check ensures that there's an alignment between what's set in the Spec and what is set in the Status.

controllers/manila_controller.go
fmount: mm  yeah, I didn't think about that: I might need an additional update to this patch to ensure we handle that case. Let me think a bit more about this and I'll come up w/ an update soon.

controllers/manila_controller.go
```diff
@@ -521,6 +530,52 @@ func (r *ManilaReconciler) reconcileNormal(ctx context.Context, instance *manila
...
+	//
+	// create NotificationBus transportURL CR and get the actual URL from the
+	// associated secret that is created
+	//
+
+	// Request TransportURL when the parameter is provided in the CR
+	// and it does not match with the existing RabbitMqClusterName
+	if instance.Spec.NotificationBusInstance != nil {
+		// init .Status.NotificationURLSecret
+		instance.Status.NotificationURLSecret = ptr.To("")
```
stuggi, line 542: I was thinking we even have to init it here and not just do the assignment later when we have the `notificationBusInstanceURL.Status.SecretName`, but its fine to do it

controllers/manila_controller.go
fmount: Thank you, I modified the `transportURLCreateOrUpdate` function to keep backward compatibility. As you suggested, passing an empty string results in the original format that we keep for regular `messagingBus` requests.
In case of `notificationBusInstance` (and actually only when we're using a dedicated rabbit), we request a new `transportURL` and we pass the rabbit instance name, used as a suffix.

tl;dr: we're able to keep the same rabbitmq instance w/o modifying it and as per my local tests it works as expected in both greenfield and brownfield scenarios.

stuggi (approved):
/lgtm

PR #442, opened 2025-07-09
Remove localtime hostMount

fmount:
`Manila` `Pods` shouldn't inherit `/etc/localtime` from the worker node/host where the `Pod` is scheduled. Mounting that volume has several drawbacks, and one of them is having these `Pods` not aligned with the rest of the control plane other than establish a dependency with the underlying host. It results difficult to track requests in case of troubleshooting and as per [1] the current approach is not recommended.
Because there is no reason to modify `localtime` to match the underlying host, this patch fixes the bug by removing it.

Jira: https://issues.redhat.com/browse/OSPRH-17676

[1] https://access.redhat.com/solutions/5487331

### DIFF
--- a/pkg/manila/volumes.go
+++ b/pkg/manila/volumes.go
@@ -23,14 +23,6 @@ func GetVolumes(name string, extraVol []manilav1.ManilaExtraVolMounts, svc []sto
 			},
 		},
 		{
-			Name: "etc-localtime",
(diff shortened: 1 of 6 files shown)


silvacarloss (approved):
/lgtm Thank you for working on this fix, Francesco!

PR #450, opened 2025-07-24
Fix notificationsBusInstance typo in the API parameter

fmount:
`Nova` and probably other operators have `notificationsBus` instead of `notificationBus`, which let storage operators to slightly diverge due to the missing final `s`.
This patch fixes the interface (currently present only in main, so it should be safe for now fixing the API parameter).


fmount:
@stuggi I think I missed this when we did the patches, sorry for that.
I spotted this typo and I created the patch to discuss with you about  fixing it for both `manila` and `cinder`.
This means we need to override `pre-commit-check` that will fail due to the `schema-crd-check`.
wdyt?

api/v1beta1/manila_types.go
```diff
@@ -150,7 +150,7 @@ type ManilaStatus struct {
 	TransportURLSecret string `json:"transportURLSecret,omitempty"`
 
 	// NotificationURLSecret - Secret containing RabbitMQ notificationURL
```
ratailor, line 152: you forgot to update in comment.

config/crd/bases/manila.openstack.org_manilas.yaml
```diff
@@ -1852,7 +1852,7 @@ spec:
                   type: integer
                 description: ReadyCounts of Manila Share instances
                 type: object
-              notificationURLSecret:
+              notificationsURLSecret:
                 description: NotificationURLSecret - Secret containing RabbitMQ notificationURL
```
ratailor, line 1856: here in description also.

api/v1beta1/manila_types.go
fmount: fixed

stuggi:
recheck

stuggi:
> @stuggi I think I missed this when we did the patches, sorry for that. I spotted this typo and I created the patch to discuss with you about fixing it for both `manila` and `cinder`. I know that we're going to inject this parameter from `openstack-operator`, so it's not a very big issue, but from a consistency point of view (e.g. if you override the parameter and you expect the same name) it might be error prone. This means we need to override `pre-commit-check` that will fail due to the `schema-crd-check`. wdyt?

yes that fix it and when other jobs are ok, we override pre-commit. need to do the same then on the openstack op bump

fmount:
> > @stuggi I think I missed this when we did the patches, sorry for that. I spotted this typo and I created the patch to discuss with you about fixing it for both `manila` and `cinder`. I know that we're going to inject this parameter from `openstack-operator`, so it's not a very big issue, but from a consistency point of view (e.g. if you override the parameter and you expect the same name) it might be error prone. This means we need to override `pre-commit-check` that will fail due to the `schema-crd-check`. wdyt?
> 
> yes that fix it and when other jobs are ok, we override pre-commit. need to do the same then on the openstack op bump

right. I'll prepare the same patch for cinder. If we land them we might be able to see the bump over the weekend in openstack-operator.

fmount:
cinder patch (for the same reason) here: https://github.com/openstack-k8s-operators/cinder-operator/pull/529

stuggi (approved):
/lgtm

PR #460, opened 2025-08-19
Use quorum queues if enabled

lmiccini:
With https://github.com/openstack-k8s-operators/infra-operator/pull/429 we added the ability to configure RabbitMQ to use quorum queues.

This commit adds a check to verify if "quorumqueues" has been added to the TransportURL secret. If its value is true, the services will be configured to use quorum queues.

Jira: https://issues.redhat.com/browse/OSPRH-19160


controllers/manila_controller.go
```diff
@@ -989,6 +989,7 @@ func (r *ManilaReconciler) generateServiceConfig(
 		return err
 	}
 	transportURLSecretData := string(transportURLSecret.Data["transport_url"])
+	transportURLQuorumQueues := string(transportURLSecret.Data["quorumqueues"]) == "true"
```
fmount, line 992: good, so this is always set to `false` unless it is set to `true` within `transportURLSecret`. I guess this is fair enough and provides backward compatibility.

test/functional/base_test.go
```diff
@@ -51,6 +51,39 @@ func CreateManilaMessageBusSecret(namespace string, name string) *corev1.Secret
 	return s
 }
 
+// CreateManilaMessageBusSecretWithQuorum creates a transport URL secret with quorum queues enabled
+func CreateManilaMessageBusSecretWithQuorum(namespace string, name string) *corev1.Secret {
```
fmount, line 55: we could put these functions to lib-common if the plan is to spread this pattern across the board. This way we can save some code and avoid duplication everywhere. For now I think I'm good with it, just something we can plan to do. Another option would be to always set `"quorumqueues"` in the existing functions (e.g. we can default to `false`), and then we can pass a parameter to set it to a different value.

```golang
// CreateManilaMessageBusSecret creates a transport URL secret without quorum queues
func CreateManilaMessageBusSecret(namespace string, name string, quorum bool) *corev1.Secret {
	return th.CreateSecret(
		types.NamespacedName{Namespace: namespace, Name: name},
		map[string][]byte{
			"transport_url": []byte(fmt.Sprintf("rabbit://%s/fake", name)),
			"quorumqueues":  []byte(string(quorum)),
		},
	)
}
```

Then you can update the signature for the existing methods to pass `false` by default, but you would avoid redefining the functions for both `Notifications` and regular `TransportURL`.
Just a starting point in case you want to improve this code now.

PR #472, opened 2025-09-19
Align and remove omitempty from databaseInstance

stuggi:
For adoption testing a ctlplane with disabled manila gets applied:

~~~
  manila:
    enabled: false
    template:
      manilaAPI: {}
      manilaScheduler: {}
      manilaShares: {}
~~~

This fails with because the manila databaseInstance parameter has omitempty specified:

~~~
$ oc apply -f ./test_deployment.yaml
The OpenStackControlPlane "openstack" is invalid:
* spec.manila.template.databaseInstance: Required value
* <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation ~~~

This aligns the databaseInstance parameter and removes the omitempty like it is speficied in all other service operators.

### DIFF
--- a/api/bases/manila.openstack.org_manilas.yaml
+++ b/api/bases/manila.openstack.org_manilas.yaml
@@ -1780,6 +1780,7 @@ spec:
                     type: string
                 type: object
             required:
+            - databaseInstance
             - manilaAPI
             - manilaScheduler
             - memcachedInstance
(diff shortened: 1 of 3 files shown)


stuggi:
seen in https://github.com/openstack-k8s-operators/openstack-operator/pull/1567

stuggi:
lets see if the depends-on works

PR #546, opened 2026-03-17
Make share v1 endpoint optional

fmount:
`v1` endpoints [are removed](https://review.opendev.org/q/topic:%22remove-v1%22+and+project:openstack/manila) from the latest manila version and not used anymore. For this reason we need to conditionally enable them based on the release. Older versions might have them enabled, and we need to preserve **brownfield** deployment for backward compatibility. However, **greenfield** deployments based on the latest version won't have them registered anymore. This patch adds a new annotation that can be used to drive the deployment and support both older and new deployment methods.

Note that by default the patch keep the existing behavior, and they are registered unless the `manila.openstack.org/sharev1: false` annotation is passed.
This will be typically done by the `openstack-operator` (and enabled in a follow up patch).

Jira: https://redhat.atlassian.net/browse/OSPRH-27947

### DIFF
--- a/api/v1beta1/manilaapi_types.go
+++ b/api/v1beta1/manilaapi_types.go
@@ -20,6 +20,7 @@ import (
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/annotations"
 
 	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -175,4 +176,20 @@ func (instance *ManilaAPI) GetLastAppliedTopology() *topologyv1.TopoRef {
 // SetLastAppliedTopology - Sets the LastAppliedTopology value in the Status
 func (instance *ManilaAPI) SetLastAppliedTopology(topologyRef *topologyv1.TopoRef) {
 	instance.Status.LastAppliedTopology = topologyRef
+}
+
+// IsShareV1Enabled - Returns false when sharev1 is explicitly disabled
+func (instance *ManilaAPI) IsShareV1Enabled() (bool, error) {
+	shareV1API, exists, err := annotations.GetBoolFromAnnotation(
+		instance.GetAnnotations(), ManilaShareV1Label)
+	if err != nil {
+		return true, err
+	}
+	// By default shareV1API is enabled to keep backward compatible with
+	// existing deployments.
+	// (Note) we can modify this if statement in 19 to reverse the behavior
+	if !exists {
+		return true, nil
+	}
+	return shareV1API, nil
 }
--- a/api/v1beta1/manilaapi_types.go
+++ b/api/v1beta1/manilaapi_types.go
@@ -17,10 +17,10 @@ limitations under the License.
 package v1beta1
 
 import (
+	"github.com/openstack-k8s-operators/lib-common/modules/common/annotations"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
-	"github.com/openstack-k8s-operators/lib-common/modules/common/annotations"
 
 	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
--- a/internal/controller/manila_controller.go
+++ b/internal/controller/manila_controller.go
@@ -1198,8 +1198,9 @@ func (r *ManilaReconciler) createHashOfInputHashes(
 func (r *ManilaReconciler) apiDeploymentCreateOrUpdate(ctx context.Context, instance *manilav1beta1.Manila) (*manilav1beta1.ManilaAPI, controllerutil.OperationResult, error) {
 	deployment := &manilav1beta1.ManilaAPI{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-api", instance.Name),
-			Namespace: instance.Namespace,
+			Name:        fmt.Sprintf("%s-api", instance.Name),
+			Namespace:   instance.Namespace,
+			Annotations: instance.GetAnnotations(),
 		},
 	}
 
--- a/internal/controller/manilaapi_controller.go
+++ b/internal/controller/manilaapi_controller.go
@@ -473,8 +473,8 @@ func (r *ManilaAPIReconciler) reconcileInit(
 		service.EndpointInternal: internalEndpointData,
 	}
 
-	apiEndpointsV1 := make(map[string]string)
 	apiEndpointsV2 := make(map[string]string)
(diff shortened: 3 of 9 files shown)


internal/controller/manila_controller.go
```diff
@@ -1195,8 +1195,9 @@ func (r *ManilaReconciler) createHashOfInputHashes(
 func (r *ManilaReconciler) apiDeploymentCreateOrUpdate(ctx context.Context, instance *manilav1beta1.Manila) (*manilav1beta1.ManilaAPI, controllerutil.OperationResult, error) {
 	deployment := &manilav1beta1.ManilaAPI{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-api", instance.Name),
-			Namespace: instance.Namespace,
+			Name:        fmt.Sprintf("%s-api", instance.Name),
+			Namespace:   instance.Namespace,
+			Annotations: instance.GetAnnotations(),
```
gouthampacha, line 1203: Hi @fmount, I assume the openstack operator will look at the openstackversions CR to determine this annotation, if yes, wouldn't we need to invoke `GetAnnotations()` in `CreateOrUpdate()` below to watch for annotation changes during an upgrade? 

I think in the brownfield case (upgrading from RHOSO 18 to RHOSO 19), the annotation should be honored and existing v1 endpoints deleted?

internal/controller/manilaapi_controller.go
gouthampacha, line 614: this is always created, even if the endpoints aren't populated.. maybe:

```
apiEndpoints := map[string]map[string]string{
      manila.ServiceNameV2: apiEndpointsV2,
  }
if v1Enabled {
      apiEndpoints[manila.ServiceName] = apiEndpointsV1
  }
```

gouthampacha, line 629:
```
if ksSvc["name"] == manila.ServiceName && !v1Enabled {
          continue
      }
```

api/v1beta1/manilaapi_types.go
```diff
@@ -20,6 +20,7 @@ import (
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/annotations"
```
gouthampacha: could go at the top since we're sorting alphabetically

internal/controller/manilaapi_controller.go
```diff
@@ -574,10 +574,7 @@ func (r *ManilaAPIReconciler) reconcileInit(
 			return ctrl.Result{}, err
 		}
 
-		// V1 (Deprected, non micro-versioned API endpoint, here for legacy users)
```
gouthampacha, line 580: nit: typo - Deprecated

gouthampacha (changes requested):
Hi @fmount - thanks for getting this in, and for checking with me!
Sorry for the delay with this review; please see some questions/comments inline

internal/controller/manilaapi_controller.go
fmount: as this block is removed, I assume you meant L594, and I fixed it, thanks!

internal/controller/manilaapi_controller.go
fmount: Technically both `apiEndpointsV1` and `apiEndpointsV2` are defined in L476 and L477, so they exist and are empty.
When a key in `apiEndpoints` is empty, the endpoint is not created (there's no value associated).
We only populate `v1` if `sharev1` is `enabled`:

```golang
if v1Enabled {
			// V1 (Deprecated, non micro-versioned API endpoint, here for legacy users)
			// will be removed when the upstream service (and dependencies) drop it
			apiEndpointsV1[string(endpointType)], err = svc.GetAPIEndpoint(
				svcOverride.EndpointURL, data.Protocol, "/v1/%(project_id)s")
....
```
The effect here is that only `v2` is created when `v2Enabled` is `false`, and we can save an extra if here as it is implicitly checked on L593.
If you like to have it explicitly defined here for `readability`, I can do that in a follow PS, but technically is not required and I only added a `Note` to make it explicit.

fmount: Thank you, it makes sense. (Done)

test/functional/manila_controller_test.go
```diff
@@ -539,9 +541,49 @@ var _ = Describe("Manila controller", func() {
...
+			mariadb.SimulateMariaDBAccountCompleted(manilaTest.ManilaDatabaseAccount)
+			th.SimulateJobSuccess(manilaTest.ManilaDBSync)
+		})
+		It("Check the resulting endpoints (sharev1 disabled)", func() {
+			// Retrieve the generated resources
+			manilaCR := GetManila(manilaTest.Instance)
+			keystoneEndpoint := keystone.GetKeystoneEndpoint(types.NamespacedName{Namespace: manilaCR.Namespace, Name: manila.ServiceNameV2})
+			endpoints := keystoneEndpoint.Spec.Endpoints
+			// sharev1 service is not created, therefore there's no "manila" endpoint key (only manilav2)
+			Expect(endpoints).ToNot(HaveKey(manila.ServiceName))
```
fmount, line 578: @gouthampacha FYI here is more explicit. We verify that when `sharev1` is disabled (see L549) no endpoint associated to `manila.ServiceName` exists, and only v2 is created.

internal/controller/manila_controller.go
fmount: Correct:
1. `OpenstackVersions CR` provide `ServiceDefaults` (which is basically the boolean associated to `sharev1`)
2. `openstack-operator`, during the reconciliation, adds the `annotation` to the `manila` `CR`. In an upgrade context, which means when the ctlplane is updated from 18 to 19 by the human operator, the boolean is switched from `false` to `true`, 
3. because of step 2 at that point you're already in the middle of the upgrade process, where manila statefulsets are rolled out. The manila controller here will simply set the annotation to the subCR (manila-api), so no need to explicitly watch for an annotation in a `CreateOrUpdate()` because annotations can't be modified by a human operator, but are hardcoded in a prefixed flow that we envision.

The result is that v1 endpoints are deleted during the upgrade reconciliation loop.
Not sure if I explained this flow properly but as a follow up for this patch we need an openstack-operator change that will enforce the annotation, and that's where we can test it.

api/v1beta1/manilaapi_types.go
fmount: Done

fmount:
Both endpoints [1] and services [2] confirm that we have no regressions with the existing jobs based on Antelope where we expect to see both v1 and v2.

[1] https://logserver.rdoproject.org/5e6/rdoproject.org/5e629a2cf2ab463085d738d4c6b1302f/controller/ci-framework-data/logs/openstack-must-gather/latest/ctlplane/endpoints

[2] https://logserver.rdoproject.org/5e6/rdoproject.org/5e629a2cf2ab463085d738d4c6b1302f/controller/ci-framework-data/logs/openstack-must-gather/latest/ctlplane/services

internal/controller/manilaapi_controller.go
gouthampacha: the note helps, thank you

internal/controller/manila_controller.go
gouthampacha: Ack, ty.. I guess I'm caught up on if the the annotation set outside the `CreateOrUpdate` will be lost on updates.. no matter if the human updates the CR or the openstack operator.. Guess we can test this flow with your changes to the openstack operator :)

gouthampacha:
/lgtm

thanks @fmount

gouthampacha (approved):

PR #561, opened 2026-04-24
Probes configuration and override support

fmount:
Add `ProbeOverrides` interface and `CreateProbeSet()` function from lib-common for `unified probe management` across `Manila` services. Enable probe customization through `CRD` overrides and remove code duplication. Updates all services (`API`, `Scheduler`, `Share`) to use the new pattern with proper scheme handling and consistent defaults.

In addition, `webhook` validation for `probes` have been introduced.

Jira: [OSPRH-29401](https://redhat.atlassian.net/browse/OSPRH-29401)


fmount:
```
  ManilaAPI (apiTimeout=60)

  ┌───────────┬─────────────────────┬─────────────────┬────────────────┐
  │   Probe   │        Field        │ Old (hardcoded) │ New (computed) │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┤
  │ Liveness  │ TimeoutSeconds      │ 10              │ 10 (same)      │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┤
  │ Liveness  │ PeriodSeconds       │ 10              │ 20             │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┤
  │ Liveness  │ InitialDelaySeconds │ 10              │ 5              │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┤
  │ Readiness │ TimeoutSeconds      │ 10              │ 10 (same)      │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┤
  │ Readiness │ PeriodSeconds       │ 10              │ 20             │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┤
  │ Readiness │ InitialDelaySeconds │ 10              │ 5              │
  └───────────┴─────────────────────┴─────────────────┴────────────────┘

  ManilaScheduler / ManilaShare (serviceDownTime=60)

  ┌──────────┬─────────────────────┬─────────────────┬────────────────┐
  │  Probe   │        Field        │ Old (hardcoded) │ New (computed) │
  ├──────────┼─────────────────────┼─────────────────┼────────────────┤
  │ Liveness │ TimeoutSeconds      │ 20              │ 10             │
  ├──────────┼─────────────────────┼─────────────────┼────────────────┤
  │ Liveness │ PeriodSeconds       │ 20              │ 20 (same)      │
  ├──────────┼─────────────────────┼─────────────────┼────────────────┤
  │ Liveness │ InitialDelaySeconds │ 10              │ 15             │
  ├──────────┼─────────────────────┼─────────────────┼────────────────┤
  │ Startup  │ TimeoutSeconds      │ 10              │ 10 (same)      │
  ├──────────┼─────────────────────┼─────────────────┼────────────────┤
  │ Startup  │ PeriodSeconds       │ 10              │ 10 (same)      │
  ├──────────┼─────────────────────┼─────────────────┼────────────────┤
  │ Startup  │ InitialDelaySeconds │ 10              │ 20             │
  ├──────────┼─────────────────────┼─────────────────┼────────────────┤
  │ Startup  │ FailureThreshold    │ 12              │ 12 (same)      │
  └──────────┴─────────────────────┴─────────────────┴────────────────┘
  ```

**Comment by fmount** [comment]
recheck

File: internal/manilascheduler/statefulset.go [hunk 1]
```diff
@@ -38,30 +38,24 @@ func StatefulSet(
...
 
-	livenessProbe := &corev1.Probe{
-		// TODO might need tuning
-		TimeoutSeconds:      20,
-		PeriodSeconds:       20,
-		InitialDelaySeconds: 10,
+	scheme := corev1.URISchemeHTTP
+	probesPort := int32(8080)
+
+	probes, err := probes.CreateProbeSet(
```
abays: I think creating a variable named `probes` here will then override further access to the `probes` package in this function.  Not necessarily a problem unless we for some reason want to access the `probes` package later on in this function in future changes.  I think this same pattern applies in the other `statefulset.go` files as well.  Not a blocker, just pointing it out.

docs/probes.md
```diff
@@ -0,0 +1,318 @@
...
+Liveness probes ensure the process is restarted if it becomes unresponsive,
+while startup probes handle the initialization window.
+
+### 5. No Startup Probes for ManilaAPI
+
+ManilaAPI starts behind HAProxy and Apache, which handle connection queuing
+during initialization. The readiness probe already prevents traffic from
+reaching the pod before it is ready to serve, making a separate startup probe
+unnecessary. The liveness probe uses a short `initialDelaySeconds` (5s) since
+the WSGI process initializes quickly.
```
abays, line 153: Should we do some sort of validation in a webhook to block the user from setting a startup probe override for `ManilaAPI`?

internal/manilascheduler/statefulset.go
fmount: Yeah, you're right, I didn't notice that but it's indeed a good point. I'm going to update the patch and provide a meaningful name that doesn't alias w/ the `probes` module.

docs/probes.md
fmount: I didn't add any validation because when I create the `manila-api` `StatefulSet` I don't assign a `StartupProbe` within the container entry. For this reason, users can provide an override for that section (`StartupProbes` is technically available), but it is not propagated to the generated `Pod`.
From a UX perspective we could add webhooks to prevent users to set it and block them in advance, but I'll probably do that in a follow up (maybe dedicated to validations) to not make this PR grow too much.

abays: Cool man, sounds good

abays (approved):
/lgtm

gouthampacha (approved):
/lgtm
/approve

thanks @fmount

PR #589, opened 2026-06-25
Set service_type in [keystone_authtoken] for access rule validation

Deydra71:
Without service_type configured, keystonemiddleware cannot validate application credentials with custom access rules, causing HTTP 401 for end users.

Closes: [OSPRH-22365](https://redhat.atlassian.net/browse/OSPRH-22365)

### DIFF
--- a/templates/manila/config/00-config.conf
+++ b/templates/manila/config/00-config.conf
@@ -45,6 +45,7 @@ user_domain_name = Default
 project_name = service
 {{- end }}
 interface = internal
+service_type = sharev2
 {{ if (index . "Region") -}}
 region_name = {{ .Region }}
 {{ end -}}


templates/manila/config/00-config.conf
```diff
@@ -45,6 +45,7 @@ user_domain_name = Default
 project_name = service
 {{- end }}
 interface = internal
+service_type = sharev2
```
fmount, line 48: Is this something we should propagate via `templateParameters`? I'm basically wondering if we should propagate [1] because for manila we have
1. sharev1: enabled in 18 but deprecated
2. sharev2: both 18 and 19, but in 19 it might have a different name and it would be good to keep the manila-operator as the single source of truth [2]

[1] https://github.com/openstack-k8s-operators/manila-operator/blob/main/internal/manila/const.go#L31
[2] https://github.com/openstack-k8s-operators/manila-operator/blob/main/internal/controller/manilaapi_controller.go#L630

fmount (approved):
/lgtm as discussed let's hardcode based on https://opendev.org/openstack/service-types-authority/src/branch/master/service-types.yaml and to keep consistency across the board.

PR #597, opened 2026-07-02
Drop privileged from manila-share StatefulSet

fmount:
Unlike some valid use cases for `glance` and `cinder`, we never require a `privileged` `Pod` for `manila-share`.
It can be always run using `Manila` `user` and `kolla` group, which is sufficient to distribute config files and start the service.

Jira: [OSPRH-32347](https://redhat.atlassian.net/browse/OSPRH-32347)

### DIFF
--- a/internal/manila/volumes.go
+++ b/internal/manila/volumes.go
@@ -81,11 +81,6 @@ func GetVolumeMounts(extraVol []manilav1.ManilaExtraVolMounts, svc []storage.Pro
 			SubPath:   "my.cnf",
 			ReadOnly:  true,
 		},
-		/*{
-			Name:      "config-data-merged",
-			MountPath: "/var/lib/config-data/merged",
-			ReadOnly:  false,
-		},*/
 	}
 
 	for _, exv := range extraVol {
--- a/internal/manilashare/volumes.go
+++ b/internal/manilashare/volumes.go
@@ -15,18 +15,8 @@ func GetVolumes(
 	propagationInstanceName string,
 ) []corev1.Volume {
 	var config0644AccessMode int32 = 0644
-	var dirOrCreate = corev1.HostPathDirectoryOrCreate
 
 	shareVolumes := []corev1.Volume{
-		{
-			Name: "var-lib-manila",
-			VolumeSource: corev1.VolumeSource{
-				HostPath: &corev1.HostPathVolumeSource{
-					Path: "/var/lib/manila",
-					Type: &dirOrCreate,
-				},
-			},
-		},
 		{
 			Name: "config-data-custom",
 			VolumeSource: corev1.VolumeSource{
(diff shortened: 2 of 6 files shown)


gouthampacha:
recheck

gouthampacha:
/lgtm

thank you @fmount! Agree. beyond the init actions, the operational parts for the share manager service (reading/writing files) doesn't need to be privileged on the host because the users have access. we might need to run privileged if we ever introduce a driver that needs to operate on a filesystem/device on the host (this could include dbus for instance). Currently, we don't support such drivers with this operator. If/when we do, we can make this conditional on the fact that someone is deploying with such a storage backend driver.

gouthampacha (approved):

fmount:
> Build failed (check pipeline). Post `recheck` (without leading slash) to rerun all jobs. Make sure the failure cause has been resolved before you rerun jobs.
> 
> https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/zuul/t/rdoproject.org/buildset/feb30f4b26dc4c8098399c99fe1b18aa
> 
> ✔️ [openstack-k8s-operators-content-provider ](https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/zuul/t/rdoproject.org/build/21519f1b08bd4d72bd9916add7eca40c) SUCCESS in 2h 05m 22s ❌ [manila-operator-kuttl ](https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/zuul/t/rdoproject.org/build/61102e1b06d54f0e9d0d3920e44859ef) **FAILURE** in 53m 05s ❌ [manila-operator-tempest ](https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/zuul/t/rdoproject.org/build/ac17ba0a071349efb91ff93535f988fa) **FAILURE** in 1h 45m 19s

@gouthampacha Not sure why this worked for me in unidelta but not in crc. I guess I have to redeploy it manually and check what's going on because I do not see logs here. Holding the patch until I reproduce and figure out the reason of the failure.

internal/manila/volumes.go
```diff
@@ -81,11 +81,6 @@ func GetVolumeMounts(extraVol []manilav1.ManilaExtraVolMounts, svc []storage.Pro
 			SubPath:   "my.cnf",
 			ReadOnly:  true,
 		},
-		/*{
```
fmount, line 84: this was a leftover

internal/manilashare/volumes.go
```diff
@@ -15,18 +15,8 @@ func GetVolumes(
 	propagationInstanceName string,
 ) []corev1.Volume {
 	var config0644AccessMode int32 = 0644
-	var dirOrCreate = corev1.HostPathDirectoryOrCreate
 
 	shareVolumes := []corev1.Volume{
-		{
```
fmount, line 21: @gouthampacha quick checking on this and looks like we do not require `/var/lib/manila` to be in the host. We might have different requirements for `lvm` in the future but for now removing this `hostPath` should allow us to not require a privileged container.

Tests on unidelta looks ok, but let's see the CI in action.

fmount: tests look good, removing the dnm label!

gouthampacha (approved):
/lgtm

PR #609, opened 2026-07-24
Remove unnecessary pods RBAC permissions

stuggi:
The workload rbacRules in manila_controller.go granted the workload service account full CRUD (create/delete/get/list/patch/update/watch) on core Pods, but that service account never reads or writes Pod
objects directly. Remove the unused rbacRules grant.

The kubebuilder RBAC markers for pods are kept (get;list), and restored on ManilaAPI, ManilaScheduler and ManilaShare where they were missing or removed: lib-common's VerifyNetworkStatusFromAnnotation,
called from those three controllers to verify NetworkAttachments, lists Pods using the controller-manager's own client, so the manager's ClusterRole (config/rbac/role.yaml) still needs get;list on pods. Regenerate config/rbac/role.yaml.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -8,7 +8,6 @@ rules:
   - ""
   resources:
   - configmaps
-  - pods
   - secrets
   - services
   verbs:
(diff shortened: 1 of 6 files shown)


abays (approved):
/lgtm

